PR #22, opened 2024-06-15
Update README.md

elwing00:
Added note about NWS weather integration


elwing00:
Closing this as the code was changed to use variables

Issue #21, opened 2024-06-13
Create Fully Kiosk JSON File Template


ADHDSquir:
With almost 400 configuration settings for Fully Kiosk, there is a lot of room for variability between setups which can cause unexpected results. We should have a standard known-good JSON config file that users can import to ease setup or to simplify troubleshooting.

ADHDSquir:
Note, Browser Mod makes the following suggesting when it detects it is running in Fully Kiosk:
![fully-screenshot](https://github.com/dinki/View-Assist/assets/1217277/75c1ffb9-37d3-47ed-b1b7-2d15300df3d6)

dinki:
Thanks again

PR #24, opened 2024-06-19
Fully config

ADHDSquir:
This is a first pass at a config JSON file that can be imported to Fully Kiosk installations. Fully has almost 400 settings, any setting not defined in a JSON file is not modified on import, so only the settings included in this file will be affected.

Addresses #21


dinki (approved):
Thanks for this. Others will definitely benefit from these settings. I'll update wiki to make this known when I get back

PR #26, opened 2024-06-20
CSS Modifications

ADHDSquir:
These changes fix some issues and may introduce others. There is still plenty more work to be done on the CSS but these commits should remove some unnecessary styles and make the core layout and clock work in either portrait or landscape. I want to get these changes in front of others for review before spending time on other views like the weather.

### DIFF
--- a/View Assist dashboard and views/views/clock/clock.yaml
+++ b/View Assist dashboard and views/views/clock/clock.yaml
@@ -20,7 +20,7 @@ styles:
     time:
       - justify-self: center
       - z-index: 2
-      - font-size: 2000%
+      - font-size: 1000%;
       - font-weight: bold
       - position: absolute
       - color: '[[[ return variables.font_color ]]]'
@@ -28,11 +28,11 @@ styles:
       - justify-self: center
       - align-self: center
       - z-index: 2
-      - font-size: 500%
+      - font-size: 250%
       - width: max-content
       - position: absolute
       - color: '[[[ return variables.font_color ]]]'
-      - top: 75%
+      - top: 65%
     night:
       - position: absolute
       - min-height: 200%
--- a/View Assist dashboard and views/views/clock/clock.yaml
+++ b/View Assist dashboard and views/views/clock/clock.yaml
@@ -13,7 +13,7 @@ styles:
         "time time"
         "date date"
         "assist assist"
-    - grid-template-rows: min-content 1fr min-content 2fr 10px
+    - grid-template-rows: '[[[ return "min-content 1fr min-content 2fr " + variables.assist_height ]]]'
   card:
     - background: '[[[ return `center / cover no-repeat url(${variables.background})` ]]]'
     - background-size: cover
--- a/View Assist dashboard and views/views/clock/clock.yaml
+++ b/View Assist dashboard and views/views/clock/clock.yaml
@@ -9,30 +9,25 @@ styles:
   grid:
     - grid-template-areas: |
         "title status"  
+        ". ."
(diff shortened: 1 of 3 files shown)


dinki (changes requested):
This is a preference thing.  My preference is to have it display slightly off the bottom.  Can you make this a variable so user can define?  Default value would be the current position.

ADHDSquir:
> Re: affix assist to bottom of screen
> 
> This is a preference thing. My preference is to have it display slightly off the bottom. Can you make this a variable so user can define? Default value would be the current position.
> 
> I'm not sure how to make comments on individual items apparently.

The commit I just pushed should handle this while being backwards compatible.

dinki:
Apologies for not being clear.  The variable should be in the dashboard configuration like the weather entity.  I do not want to add this as a device specific setting.  The user can choose one setting to use across all devices for that dashboard.

ADHDSquir:
> Apologies for not being clear. The variable should be in the dashboard configuration like the weather entity. I do not want to add this as a device specific setting. The user can choose one setting to use across all devices for that dashboard.

Ah, that makes more sense. Moved it.

View Assist dashboard and views/views/clock/clock.yaml
```diff
@@ -55,19 +52,21 @@ custom_fields:
...
+          - line-height: 1
+          - color: '[[[ return variables.font_color ]]]'
         icon:
-          - width: 80%
-          - color: white
+          - width: 10vmin
+          - color: '[[[ return variables.font_color ]]]'
+          - height: 60%
+          - margin: 0 1vmin 40% 0
+  time: '[[[ return variables.var_current_time ]]]'
```
jlmaners, line 69: Either line 69 or 70 needs to be removed as it is the same declaration and will cause the view to throw an error when trying to edit the card in the HA UI.

View Assist device configuration/device_config_example.yaml
```diff
@@ -39,4 +39,4 @@ template:
         alarm: "idle"
         cycle_view: ""
         do_not_disturb: false
-        
+
```
jlmaners, line 42: Should consider removing this commit from the PR as it appears to be an extraneous line change

jlmaners (changes requested):
I tested this and other than a one line issue in clock.yaml this looks good and works to make the display handle portrait and landscape better.

dinki:
@ADHDSquir 
Thanks for your work on this, it looks really great on the TSV and I'm happy to see that the dashboard icons on top right don't cause the weather data to move up and down.  Was that done by stretching the "title status" to two rows?  If so, I think this modification should be made in the dashboard so that all views can benefit.  Again, I am no CSS expert but it would be great to adjust the actual height of that row rather than add an additional row as that appears to be shifting the clock down a bit and it is appearing off center.

The view also looks great in portrait on the TSV and my phone.  Landscape on the phone does not show the date at all.  Can you try on your phone and see if you see the same?  

Can you test that the tap on weather navigates to the weather page?  For unknown reasons mine is not responding to the tap.  I just set this new dev environment up so it may just be a me problem.


I think the work in this PR is definitely a step in the right direction.  Would you mind considering making these changes and  the changes @jlmaners mention above in regards to the time declaration twice and the unnecessary change to the device configuration file?

Again, many thanks on your work on cleaning up my mess.

dinki:
Thanks for your submission and your patience.  Can you please submit this to branch viewassist-dashboard and we will work on things there.  Accepting to main with all these changes could pose problems.

PR #34, opened 2024-08-07
Add automation to use MQTT to change brightness based on mode

dovy6:


dinki:
Thank you for your submission.  I am running almost the exact same thing here with great results.

I have been thinking of how to handle things like this.  As you know, VA is mostly a framework with lots of options and choices for users.  As such, about half the folks I've spoken to use Fully Kiosk and the other half use Wallpanel.  Further, not everyone uses MQTT like we do.

In short, it's hard to include something like this as it doesn't have 'universal' support and has dependencies outside of the core requirements.  In the same sense, though, it does provide value for those that do.  I am open to discussing how to best present things like this.  Maybe in a subcategory for wallpanel specific automations?  Even so, this would really need to go in the main control automation which makes things harder to maintain multiple versions.  

Really I'm trying to prevent confusion, calls for added support, and fracturing things to the point of problems maintaining.  Not exactly sure how to handle this and I know this is just the first of many like cases so will certainly need to think and discuss.   Please don't take this as a slight to your efforts.  I certainly appreciate it but I am finding myself spending more time supporting than I am creating new functions which is not a great thing.

dovy6:
Yeah, I totally get where you are coming from. I was thinking of taking some time and separating VA into "core" and "non-core" sections. For example, the control automation, setting up the VA devices themselves, seem like core functionality. Some of the dashboards, the custom sentences, etc, seem like none core, optional setup features. This would simplify setup for those new to the project - do these x steps and VA is set up, while also allowing the power of the project to still be there - you can also set up AI, custom sentences, etc etc.

What do you think about that?

Dovy

dovy6:
Well I tried to created a new PR with the files organized, but I think instead it modified this one? 

Also, it seems that the wiki does not come over when I forked the VA project. I would like to work on the documentation a bit too. Perhaps the wiki could be organized int a /docs folder?

dinki:
May need to close this one and open a new one.  In honesty, I am not super versed in GH so I'm learning as I go.

We are working on rolling a new wiki and we will hopefully have a separate repo for it so that folks who only want to work on doc updates can do so easily.  All a work in progress.  :)

PR #35, opened 2024-08-08
Complete reorganization of folder structure

dovy6:
Feel free to agree or disagree with exactly what's core or non core and change appropriately.


dinki:
Thank you for your submission.  I do appreciate it but I am not prepared to make a change of this magnitude.  This will break every blueprint link, every wiki link and every youtube link.  I'm not prepared to make that sweeping change at this time.

I think this will more than likely be handled through the wiki in the short term and will perhaps be reorganized at some later date.  I appreciate your efforts but this is way too much work for me at this time.

PR #37, opened 2024-08-08
Update blueprint-whattimeisit.yaml

Flight-Lab:
Fix variable "view" to use the user selected input view and not the hard coded /dashboard-viewassist/clock


dinki:
Thank you for the fix.  Much appreciated.

PR #46, opened 2024-08-14
Fix typo in webpage BP

dovy6:


dovy6:
Believe this should be fix typo only, no brightness automation in this fork. Did I do it correctly?

PR #47, opened 2024-08-17
Currency Converter international

AndrewSteel:
Internationalization of currency names. Extension of the Blueprint and Python scripts.


dinki:
Sorry but this has a lot of changes to files outside of Currency Convertor so I won't be able to merge.  I am new to GH so maybe something can be done on my side but not sure.  

What you could do, if willing, would be to check out again to a new branch on your repo and then just copy over just those changed files and make the PR again.  Apologies again but this could make a mess of things once this goes to man as a PR.

PR #51, opened 2024-08-19
Update blueprint-howstheweather.yaml

Flight-Lab:


dinki:
```
default: "(How's | How is | What's | What is) the weather"
```
This would allow user to say 'the weather' which to me is odd.  Now perhaps just 'weather' would be good:
```
default: "(How's | How is | What's | What is) (the) weather"
```

Thoughts?

dinki:
I'll accept and make the change right quick.  Thanks.

PR #52, opened 2024-08-19
Update spellaword.yaml

Flight-Lab:


dinki:
I've had trouble with these before.  I think HA has problems with nested parenthesis which this one does not have.  Want to be certain that you tested these options before pushing this one.  Thanks.

dinki:
So this one is confirmed working?

dinki:
Thanks for your continued effort to make View Assist better!

PR #54, opened 2024-08-19
Update blueprint-viewcamera.yaml

Flight-Lab:


dinki:
Thanks for the submissions.  Have you verified that all of these work and coexist with other custom sentences?  I have made it a point to make these mostly user definable in a way similar to what you provided.  This makes for an easy time for users to do what they did and to also translate it into their language of choosing.

I am unsure that I want to change these for that reason but will look through all of them to see if I might accept some.  You are on discord right?  I wouldn't mind chatting about these there if so.  Your GH name doesn't match discord username, right?

PR #57, opened 2024-08-20
change currency filename

AndrewSteel:
Changed currency filenames, adapted pyscript file and changed data section in blueprint. The old version didn't work for me.


dinki:
Looks good.  Pulling into test branch for, well, testing.  Will report back here and/or Discord.  Thanks Andrew!

dinki:
@AndrewSteel Can you confirm that pyscript is still working?  I'm getting back -1 in the return.

AndrewSteel:
Oh yes --- "error": -1, "message": "Unknown currency".
The error occurs when one of the two currencies cannot be resolved. In my
case, the error occurred because the TTS component returned "US Dollar" on
one occasion and "US-Dollar" on another. The error also occurs if, for
example, "EUR" or "USD" is entered as the base currency in the blueprint
instead of "Euro" and "US Dollar". However, this blueprint error will be
fixed in the new version of the Currency Convertor.
Please also note that the localization is almost finished. It will be
implemented with the Currency Convertor version as an example and can be
used in all other ViewAssist automations in the future.
This will remove the language-specific parts of the blueprints for both
speech input and output, speech-text and on-screen text. In addition, the
automation can be easily adapted to a new language without changing the
blueprint or an associated Python script.
I still need to make some adjustments to the blueprint and the
currencyconvertor.py-script. And I am writing a concept paper about it. I
hope to have it done before my vacation next week and you can find it on my
github.


Am Mi., 21. Aug. 2024 um 21:59 Uhr schrieb Donny F ***@***.***

dinki:
Sounds great.  I'll hold off until I hear back.  Thanks Andrew!

PR #60, opened 2024-08-21
Update blueprint-thermostatcontrol.yaml

elwing00:
took out hardcoded thermostat and added some more options for commands


dinki:
Thank you for this.  I must have been half asleep to miss that hardcoded variable!

PR #61, opened 2024-08-21
Alternative Music View

jsittner:
Here is an alternative music view.  The view requires [mini-media-player](https://github.com/kalkih/mini-media-player).

image


dinki:
Accepted and thanks again!

PR #64, opened 2024-09-06
Navigate to Music View & Set Music Mode When Music Playback Begins

Flight-Lab:
Currently, the only way into and out of the music view and music mode are through voice controls in the play music and device functions automations. 

This PR navigates to the music view and sets music mode when a View Assist Satellite musicplayer_device has begun playing, regardless of how playback was activated. This is convenient when activating music playback from a separate device, e.g., starting playback with your phone or when syncing a media player to an already playing player or speaker group.

Please, your suggestions, adjustments, or reasons not to include are most welcome.

### DIFF
--- a/View Assist custom sentences/Device Functions/blueprint-devicefunctions.yaml
+++ b/View Assist custom sentences/Device Functions/blueprint-devicefunctions.yaml
@@ -228,15 +228,15 @@ action:
             id:
               - stop_music
         sequence:
-          - data:
-              entity_id: "{{ target_satellite_device }}"
-              mode: normal
-            action: python_script.set_state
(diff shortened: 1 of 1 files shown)


dinki:
Thanks for your work here.  This may be problematic as I believe the broadcast blueprint will also trigger this as Chime TTS uses the media player to play what I believe is just an audio file.  Can you test this and confirm that it does not trigger this screen change?

dinki:
Also this appears to contain the same device functions improvements from the other PR.  That one has been accepted but would like this PR to be clean so please remove that file if it does not belong to this PR.  Thanks.

PR #65, opened 2024-09-06
Fix music stop command blueprint-devicefunctions.yaml

Flight-Lab:
Fix music stop command.
Move media_player.stop action to be first in sequence.


dinki:
Thanks for your continued support.

PR #70, opened 2024-09-11
Update microphone.md

droyholmes:
Fixed order of operations in IPWebcam Section, also changed wording to express end state of indicator instead of assuming the user just has the same settings and telling them to "Tap" this can make troubleshooting later more difficult

### DIFF
--- a/wiki/docs/tablet-satellite/android-config/microphone.md
+++ b/wiki/docs/tablet-satellite/android-config/microphone.md
@@ -23,9 +23,9 @@ Detailed install video: https://youtu.be/-7cHfqZMa1c
 
 ### Configuration
 1. Find the IPWebcam Icon and tap it to launch.
-2. Scroll down until you see the **Service Control** heading and tap on **Optional Permissions**.
+1. Scroll down until you see the **Service Control** heading and tap on **Optional Permissions**.
 1. Tap **Allow streaming in background** and then turn on **Allow display over other apps**.
-1. Tap back and then activate the checkbox on **Show camera overlay**. This will prevent another window from appearing over your dashboard.
+1. Tap back and then deactivate the checkbox on **Show camera overlay**. 
 1. Tap back and scroll up until you see **Power management** tap it and under *Misc* Turn on **Disable notification** and **Stream on device boot**. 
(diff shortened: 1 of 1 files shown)


wiki/docs/tablet-satellite/android-config/microphone.md
```diff
@@ -23,18 +23,18 @@ Detailed install video: https://youtu.be/-7cHfqZMa1c
 
 ### Configuration
 1. Find the IPWebcam Icon and tap it to launch.
-1. Scroll down until you see the *IP Webcam settings* section.
-1. Tap on **Power management** and under *Misc* Turn on **Disable notification** and **Stream on device boot**. 
+2. Scroll down until you see the **Service Control** heading and tap on **Optional Permissions**.
```
dinki, line 26: This should start with '1.' and not '2.'

wiki/docs/tablet-satellite/android-config/microphone.md
```diff
@@ -23,18 +23,18 @@ Detailed install video: https://youtu.be/-7cHfqZMa1c
...
+1. Tap **Allow streaming in background** and then turn on **Allow display over other apps**.
+1. Tap back and then activate the checkbox on **Show camera overlay**. This will prevent another window from appearing over your dashboard.
+1. Tap back and scroll up until you see **Power management** tap it and under *Misc* Turn on **Disable notification** and **Stream on device boot**.
```
dinki, line 29: Why not put this closer to start so you don't scroll down then back up?  I don't see where order matters much on these as long as the settings are set.  Is there a reason?

dinki:
Couple things I've noticed.  You've got a '2.' that should be '1.' .  Also I am looking through my settings and note that I do not have:
```
1. Tap back and then activate the checkbox on **Show camera overlay**. This will prevent another window from appearing over your dashboard.
```
Set.  I THINK this the thing that causes WebIPCam to have the persistent notification.  Did you mention you were seeing this?  I may opt to remove this but I will change.

Thanks for your efforts to make View Assist better.  I'll approve once the two comments I made are addressed.

dinki (approved):

dinki (approved):

PR #71, opened 2024-09-17
Fix dust and tornado weather icons in dashboard.yaml

Flight-Lab:
Missing spaces were added


dinki:
Thanks for your continued efforts to make the project better!

PR #72, opened 2024-09-25
Update input description blueprint-devicefunctions.yaml

Flight-Lab:
Removed duplicate word in input description.
&
"Levels" changed to "level" to maintain consistency.

Both "levels" and "level" were used. Which do you prefer?

"Level" makes the most sense to me, as the volume can only be set to one level at a time.


dinki:
Yep.  Reads better your way.  Thanks!

PR #73, opened 2024-09-25
more details in conversation response for set mode blueprint-devicefunctions.yaml

Flight-Lab:
When a nonexistent mode is called, the conversation response will now list the different mode options as they appear in the modes input field.


dinki:
Nice addition!  Thanks for your continued submissions.

PR #77, opened 2024-10-02
Update m5stack-atom-echo.md

TohurTV:
Added guide for using M5Stack Atom Echo as microphone for View Assist satellites


dinki:
Thanks again for taking the time to put this together!

PR #79, opened 2024-10-02
Visible HA Timer

Flight-Lab:
This is just a draft to get all the changes in one place for a better discussion.

**viewassist-timer.py has new service named "get_time_difference2".**
Open to name suggestions.
"get_time_difference2" does the same thing as "get_time_difference" but returns the time difference in HH:MM:SS format to work with the timer.start action.

**blueorint-alarmsreminderstimers.yaml has...**
New inputs:
timer_view (need to create a new view)
user_timer

New trigger

New Variable

New choose option

Creates the ability to start an HA based timer and view that timer on a new view

**VA device config file gets new attribute**
`user_timer: timer.(name)`
Name is work in progress, open to suggestions. Should probably end in _device to maintain consistency.
must also create new timer

**New "Timer" view**
[Uses Circular Timer Card](https://github.com/karlis-vagalis/circular-timer-card)

This is all just a proof of concept to start a discussion on how best to implement the timer.

### DIFF
--- a/View Assist device configuration/device_config_example.yaml
+++ b/View Assist device configuration/device_config_example.yaml
@@ -30,6 +30,7 @@ template:
         musicplayer_device: "media_player.browsermod-livingroom_2"
         display_device: "sensor.browsermod-livingroom_browser_path"
         browser_id: "ViewAssist-livingroom"
+        user_timer: "timer.viewassist_livingroom_user_timer"
         timer_device: "timer.viewassist-livingroom"
         view_timeout: "20"
         mode: "normal"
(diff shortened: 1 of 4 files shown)


dinki:
When you can, please update this file leaving the header as is:
https://github.com/dinki/View-Assist/blob/main/wiki/docs/community-contributions/cc-sentences/extended-device-controls.md
dinki — Today at 2:30 PM
Oops.  Looks like you already provided.  Can you cancel this PR and put that data in the file mentioned above?
https://github.com/dinki/View-Assist/pull/81

----

This is from GH.  Can you please submit using the file mentioned?  Thank you!

dinki:
So I really like the idea of having an actual timer running to display these values as  true countdowns but I also dislike, very much, creating additional helpers which I think this would require.  I'm hesitant to request the user create these additional helpers especially being per device.  I see the utility but not sure that it is a need as this information can be obtained as a TTS reply to the voice request.

It's very frustrating that HA does not provide 'disposable' timers.  Perhaps we will see the VA devices become 'compatible' devices with HA's built in timers and then something like this becomes much easier.

Apologies and thanks for your understanding.  It's not easy making some of these decisions but I do think waiting is the right thing to do.

PR #80, opened 2024-10-04
Update blueprint-traveltimesbywaze.yaml

elwing00:
Added Source URL to be able to update/import


dinki:
Doesn't this happen automatically?   I did recently have to update all BP install links to get rid of the spaces that were breaking things but I THOUGHT that once a user installs these it will allow for updating without the need to add this line?

elwing00:
It wouldn't let me, said "no source_url" as the error.  All the other blueprints have a source url in them.

dinki:
I did this search and did not find anything like that in the repo:

https://github.com/search?q=repo%3Adinki%2FView-Assist+source&type=code

Now the local copies in HA will have that value but it is inserted on install.

image

elwing00:
maybe just my install - feel free to reject the PR

PR #89, opened 2024-10-31
Currency Convertor with new localization

AndrewSteel:
I have checked the German phrases and they will work.


dinki:
Thank you again for your thorough work on this.  I will merge in now and give it a test today or tomorrow.  I may have additional questions but more than likely not.  You are a superstar!

PR #90, opened 2024-11-10
Update music-alternative.yaml - entity variable fix

Flight-Lab:
Fix entity variable used by mini-media-player.


dinki:
thanks for this

PR #91, opened 2024-11-10
Update blueprint-devicecontrol.yaml - fix error when saving w/ no mic…

Flight-Lab:
Use variable micdevice instead of !input micdevice.

This fixes the error preventing the automation from being saved when no mic device is applied in the inputs.

### DIFF
--- a/View_Assist_control_automations/blueprint-devicecontrol.yaml
+++ b/View_Assist_control_automations/blueprint-devicecontrol.yaml
@@ -329,7 +329,7 @@ action:
       then:
       - service: switch.turn_on
         target:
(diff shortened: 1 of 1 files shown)


dinki:
I think that will get you past but the better way to do it is to make the default an empty list.  I'll give this a go first and then circle back here.  I'm not sure what will happen if the user does set this value and need to use it if the !input definition hasn't pulled it.

https://community.home-assistant.io/t/blueprint-optional-input/278810/2?u=dinki

dinki:
Closing this one as another fix is available.  Thanks for looking into it.

PR #92, opened 2024-11-10
Dashboard 2.0 various fixes

Flight-Lab:
Dust and tornado were missing a space preventing them from being included.

Replace hard-coded mic device with variable mic device.


dinki:
Nice!  Yea I missed that hardcoded value when testing.  Thank you!

PR #98, opened 2024-11-17
Assist Bar Updates to dashboard.yaml

jsittner:
- Updated the blur pop up and flashing bar with some better logic.
- Added listening and processing colors to the flashing bar.


dinki:
I think we got this straight on Discord.

PR #101, opened 2024-12-03
Beta testing fixes

michelle-avery:
Fixed a few typos I found while testing. Although it would be _really_ interesting to use my dog as a mic device, it looks from discord like that was a troubleshooting step.


dinki:
I sure thought I had fixed that.  Thank you!

PR #105, opened 2024-12-15
RtpMic is not available on googles play store, might want to add the …

mannebk79:
…last version to a project owned location and remove google playstore link


dinki:
Thanks.  RTPMic and WebIPCam will soon be replaced with the suggestion to use HassMic in the very near future.

PR #106, opened 2024-12-21
Update blueprint-playmusicwithmusicassistant.yaml

jeroenterheerdt:
Music assistant is moving to HA Core and the service names are updating from mass to music_assistant. Search is not available yet, but play_media is.


dinki:
Thanks for putting this in.  I had planned to do this when search was included which I think will be for 2025.1 in a few weeks.  The existing BP works for me currently with the HACS extension so I hesitate in making any changes until search is available.  I'll push this PR when that happens.  

Thanks for saving me this effort for sure!

dinki:
I chatted with the MA developer last night and he told me that music_assistant.search will be available today in beta and will definitely be part of the 2025.1 release.  Will be pushing this one soon based on that.

dinki:
These are now available in 2025.1 so pushing through.  Thanks again!

PR #107, opened 2025-01-02
Clock Webpage Background

janstadt:
adding an example of how to use a webpage slideshow as the background of the clock


dinki:
Looks great!  Thanks for your contribution!

PR #109, opened 2025-01-05
Update blueprint-devicecontrol.yaml - fix select_random_image action

brettmiller:
action should be python_script.select_random_image not pyscript.select_random_image

### DIFF
--- a/View_Assist_control_automations/blueprint-devicecontrol.yaml
+++ b/View_Assist_control_automations/blueprint-devicecontrol.yaml
@@ -612,7 +612,7 @@ action:
       - condition: template
         value_template: "{% if rotate_background %}true{% endif %}"          
     sequence:
-      - action: pyscript.select_random_image
+      - action: python_script.select_random_image
         data: 
           directory: "{{ background_directory }}"
         response_variable: random_image


dinki:
@brettmiller Not sure if you were having problems before but it definitely is pyscript.select_random_image

image


This file should be put in the pyscript directory.  Did you put it in python_scripts dir?  If so, that might be where the disconnect is.  Hopefully I was explicit in the wiki but perhaps not.  Let me know.  Thanks!

brettmiller:
Ah ok. I'm just coming back to working on this after starting months ago and was mostly just going back through the official doc and trying to supplement with https://github.com/dinki/View-Assist/tree/main/View%20Assist%20dashboard%20and%20views. The 2.0 page doesn't really make it clear where the pyscript.select_random_image should go, so I assumed it should go with the other script (set_state.py) in `config/python_scripts`. I also hadn't seen a mention of `pyscript` anywhere else, but searching it looks like it gets mentioned directly in several of the individual extended sentence Readme files, but I hadn't read any of those yet.

dinki:
Great feedback.  Thanks for pointing this out.  I'll make this clearer.

dinki:
Wiki updated.  Let me know if you run into any other issues.

brettmiller:
Also of note, the Pyscript integration needs to have the Allow All Imports option set otherwise it can't load the viewassist-select_random_image.py script. It fails at `import os`.

dinki:
I do have a YT video on pyscript but probably need to make a wiki page for it too.

PR #113, opened 2025-01-11
Update blueprint-broadcast.yaml

jeroenterheerdt:
simple fix to make broadcast work again. broadcast failed for me because `group_entity` was not set when `target_satellite_device` was set, reversing the order fixed it for me.


dinki:
Sheez.  At one point this all worked.  Can't say what happened to cause this but probably code clean up without thinking that order mattered.  Thanks for the fix!

PR #114, opened 2025-01-11
Create full-night-mode

jeroenterheerdt:
Add a full-night-mode md to describe what I did on my satellite to have them turn into "night mode" showing different apps. One of my satellite shows a night clock app and another turns its screen off until reactivated in the morning.

### DIFF
--- a/wiki/docs/community-contributions/cc-tipsandtricks/full-night-mode
+++ b/wiki/docs/community-contributions/cc-tipsandtricks/full-night-mode
@@ -1,0 +1,51 @@
+---
+title: Full night mode
+sidebar_position: 2
+---
+​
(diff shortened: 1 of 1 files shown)


jeroenterheerdt:
wait a minute, I think we can also switch apps directly through webpanel, so if you just wanted to turn off the screen you don't even need the home assistant app. Investigating and will update

jeroenterheerdt:
yep, it can be done, but I still prefer the Home Assistant app way so the View Assist dashboard is shown consistenly. Anyway, added that option to the description as well.

dinki:
A perfect community contribution!  Thanks again for your continued additions and improvements

PR #116, opened 2025-01-15
Further HassMic Support for Extended Device Controls

Flight-Lab:
Sets wake alert sound and STT error sound to play with announce: true


dinki:
looks good to me

PR #117, opened 2025-01-16
Update lenovo-thinksmart-view.md with extra info on the webview version.

jeroenterheerdt:
As discussed on Discord. We should also maybe consider adding a "FAQ" or "troubleshooting" section somewhere in the docs. I think this webview question might come up a couple of times since it's easy to miss. Happy to add the info in there, but I'd prefer if you decide where this can go.


dinki:
Thanks for this.  Yes, agree to the need to compile and present a FAQ.

jeroenterheerdt:
> Thanks for this. Yes, agree to the need to compile and present a FAQ.

any idea on where you want the FAQ to go? Happy to put something together so we can start compiling the FAQs.

dinki:
image

How about here?

jeroenterheerdt:
consider it done

jeroenterheerdt:
I screwed up my fork, so I'll wait until you close my other to PRs so it's less messy before making the PR for the FAQ.

PR #118, opened 2025-01-16
Update blueprint-playmusicwithmusicassistant.yaml

jeroenterheerdt:
Bunch of stuff here (untested due to nighttime here, so will likely see more commits before ready to merge!):

- more sentences
- volume control (up and down)
- playback (pause/resume/next song)
- response feedback on success and failure
- overall restructuring of the code to make it easier to maintain.

Todo:
- test, test, test!
- Ideally I'd like to make this more elegant by removing the duplication, but this is already a step forward (I hope).

### DIFF
--- a/View_Assist_custom_sentences/Play_Music_with_Music_Assistant/blueprint-playmusicwithmusicassistant.yaml
+++ b/View_Assist_custom_sentences/Play_Music_with_Music_Assistant/blueprint-playmusicwithmusicassistant.yaml
@@ -1,6 +1,6 @@
 blueprint:
   name: View Assist - Play Music with Music Assistant
-  description: Use various methods to play individual songs, add to queue, and play artist from your Music Assistant server (View Assist Play Music with Music Assistant v 1.0.3)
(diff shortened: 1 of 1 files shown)


dinki:
Thank you for this.  You've done some things that have been on my want list for a while.  A few things to note:

* `media_player.picoreplayer` is defined by mistake
* The media player controls for volume, play, etc should be put in the [device functions](https://github.com/dinki/View-Assist/tree/main/View_Assist_custom_sentences/Device_Functions) blueprint .  Some functions are already there but you definitely have new ones to add.  Please move those there and submit a separate PR

I have been thinking about multilingual support for the blueprints.  Unfortunately it is not easy.  I was thinking about adding the commands as inputs like I have on other blueprints ([example](https://github.com/dinki/View-Assist/blob/main/View_Assist_custom_sentences/Hows_the_weather/blueprint-howstheweather.yaml#L7-L11)) so that people can change these to fit their languages.  Thoughts on that?

Thanks again!

jeroenterheerdt:
going to close this and replacing with new PR, due to clean-up of my fork.

PR #125, opened 2025-01-17
Update blueprint-devicecontrol.yaml FOR TALKING AVATARS IMPLEMENTATION

relust:
These are the changes to the device control blueprint code that I made to implement my Talking Avatars project. I introduced the two gifs for speech and for listening in the graphic interface. I have not added the automations for custom responses and continue conversation, but I will put their blueprints separately, so that after they are tested and adapted they can be integrated into this blueprint.

### DIFF
--- a/View_Assist_control_automations/blueprint-devicecontrol.yaml
+++ b/View_Assist_control_automations/blueprint-devicecontrol.yaml
@@ -105,7 +105,24 @@ blueprint:
               custom_value: true
               options:
                 - flashing bar
-                - blur pop up         
+                - blur pop up
+                - avatar
+        speech_gif:
+          name: Avatar speech gif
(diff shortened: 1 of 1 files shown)


dinki:
As mentioned in chat, I will look/try this next week.  One thing in the meantime though.  I would like you create another section at the bottom labeled as `Optional.  Talking Icon specific` and move this input code there:
https://github.com/dinki/View-Assist/blob/40dc1d157ab4e6da1dd0cdf5b37c619eff05d468/View_Assist_control_automations/blueprint-devicecontrol.yaml#L110-L125

Follow the same style as what is shown for the stream assist optional sectioning.

dinki:
Closing this for now.  Will need to figure out how this can be added to the integration.

PR #126, opened 2025-01-18
first version of updated blueprint for MA

jeroenterheerdt:
Replaces #118.

Original comment:

Bunch of stuff here (untested due to nighttime here, so will likely see more commits before ready to merge!):

more sentences
volume control (up and down)
playback (pause/resume/next song)
response feedback on success and failure
overall restructuring of the code to make it easier to maintain.
Todo:

test, test, test!
Ideally I'd like to make this more elegant by removing the duplication, but this is already a step forward (I hope).

Comment from @dink which still needs to be taken into account:
Thank you for this. You've done some things that have been on my want list for a while. A few things to note:

media_player.picoreplayer is defined by mistake
The media player controls for volume, play, etc should be put in the [device functions](https://github.com/dinki/View-Assist/tree/main/View_Assist_custom_sentences/Device_Functions) blueprint . Some functions are already there but you definitely have new ones to add. Please move those there and submit a separate PR
I have been thinking about multilingual support for the blueprints. Unfortunately it is not easy. I was thinking about adding the commands as inputs like I have on other blueprints ([example](https://github.com/dinki/View-Assist/blob/main/View_Assist_custom_sentences/Hows_the_weather/blueprint-howstheweather.yaml#L7-L11)) so that people can change these to fit their languages. Thoughts on that?

Thanks again!

### DIFF
--- a/View_Assist_custom_sentences/Play_Music_with_Music_Assistant/blueprint-playmusicwithmusicassistant.yaml
+++ b/View_Assist_custom_sentences/Play_Music_with_Music_Assistant/blueprint-playmusicwithmusicassistant.yaml
@@ -1,6 +1,6 @@
 blueprint:
   name: View Assist - Play Music with Music Assistant
-  description: Use various methods to play individual songs, add to queue, and play artist from your Music Assistant server (View Assist Play Music with Music Assistant v 1.0.3)
+  description: Use various methods to control volume and playback and play individual songs, add to queue, play artist and playlists from your Music Assistant server (View Assist Play Music with Music Assistant v 1.0.3)
   domain: automation
   input:
     group_entity:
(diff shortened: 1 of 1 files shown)


View_Assist_custom_sentences/Play_Music_with_Music_Assistant/blueprint-playmusicwithmusicassistant.yaml
```diff
@@ -24,153 +24,262 @@ blueprint:
...
-            target:
-              device_id: "{{ target_display_device }}"        
-          - service: music_assistant.search
-            data:
+          - data:
               limit: 1
               config_entry_id: "{{ ma_instance }}"
               media_type:
-                - artist
+                - "{{trigger.id}}"
```
dinki, line 87: I'm not sure that this is correct.  Why should this not be artist?

View_Assist_custom_sentences/Play_Music_with_Music_Assistant/blueprint-playmusicwithmusicassistant.yaml
```diff
@@ -24,153 +24,262 @@ blueprint:
...
+              - metadata: {}
+                data:
+                  path: "{{ view_path }}"
+                target:
+                  device_id: "{{ target_display_device }}"
+                action: browser_mod.navigate
+              - target:
+                  entity_id: "{{ target_musicplayer_device }}"
+                data:
+                  media_type: "{{trigger.id}}"
```
dinki, line 111: same question as above.  Does this work?

View_Assist_custom_sentences/Play_Music_with_Music_Assistant/blueprint-playmusicwithmusicassistant.yaml
jeroenterheerdt: This is correct and does work. It basically refuses the trigger ID to pass the right context to MA. I did this so it's easier to add other searches in the future.

View_Assist_custom_sentences/Play_Music_with_Music_Assistant/blueprint-playmusicwithmusicassistant.yaml
jeroenterheerdt: Yes.

View_Assist_custom_sentences/Play_Music_with_Music_Assistant/blueprint-playmusicwithmusicassistant.yaml
dinki: I see what you are doing.  I think these are unnecessary changes.  Please revert all "{{trigger.id}}" changes back to original.

jeroenterheerdt:
These are best practices changes but if that's what you want I'll change it back.

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

View_Assist_custom_sentences/Play_Music_with_Music_Assistant/blueprint-playmusicwithmusicassistant.yaml
```diff
@@ -1,6 +1,6 @@
 blueprint:
   name: View Assist - Play Music with Music Assistant
-  description: Use various methods to play individual songs, add to queue, and play artist from your Music Assistant server (View Assist Play Music with Music Assistant v 1.0.3)
+  description: Use various methods to control volume and playback and play individual songs, add to queue, play artist and playlists from your Music Assistant server (View Assist Play Music with Music Assistant v 1.0.3)
```
dinki, line 3: `control volume and playback and` should be removed as it was moved to device-function BP

View_Assist_custom_sentences/Play_Music_with_Music_Assistant/blueprint-playmusicwithmusicassistant.yaml
```diff
@@ -24,156 +24,192 @@ blueprint:
       description: The View Assist dashboard view to use for displaying music information (example /dashboard-viewassist/music)
       default: "/dashboard-viewassist/music"
 alias: View Assist - Play Music with Music Assistant
-description: ""
-trigger:
-  - platform: conversation
-    command:
+description: "Control volume, playback and play music using music assistant"
```
dinki, line 27: Reword to not include control volume

View_Assist_custom_sentences/Play_Music_with_Music_Assistant/blueprint-playmusicwithmusicassistant.yaml
```diff
@@ -24,156 +24,192 @@ blueprint:
...
+                  by {{trigger.slots.artist}} now!
+            else:
+              - set_conversation_response: >-
+                  I was unable to find any song called {{trigger.slots.song}} by
+                  {{ trigger.slots.artist }}
 variables:
-  view_path: !input view_path
-  group_entity: !input group_entity
-  ma_instance: !input ma_instance
+  view_path: /dashboard-viewassist/music
```
dinki, line 210: please revert this and the next two as they are inadvertently set to your specifics.

dinki:
Made a few comments but will still need to try this.  Sorry for the delay.

View_Assist_custom_sentences/Play_Music_with_Music_Assistant/blueprint-playmusicwithmusicassistant.yaml
jeroenterheerdt: also removed control here as it's not in this blueprint either as was moved to the control BP.

dinki:
Trying it now.  Running into problems:

"ok nabu, Play Banana Pancakes by Jack Johnson"
"Done"

image

"ok nabu, play some Jack Johnson"
Does not match intent.  Breaking change.  You have removed how the sentence could be called previously.  Please revert or include

"ok nabu, play some Jack Johnson music"
"Playing some Jack Johnson music quote"
Music does not start.

Do you have Spotify?  If so, can you try with these examples?

jeroenterheerdt:
I don't have spotify so can't test this exact utterances but will double check that they work!

dinki:
Understood and no problem.  Just don't want breaking changes.  

I really should have tried the original automation before reviewing yours.  I think the:

"ok nabu, play some Jack Johnson music"
"Playing some Jack Johnson music quote"
Music does not start.

Would have played.  Seems like a speaker problem on my side as when I switched back it did the same thing.  The one with screenshot did give an error though so it will need more work.

Still hold to the request to keep things the same and add functionality unless you find something glaring that is wrong though.

Thanks for the continued suport and submissions.  It is appreciated.

jeroenterheerdt:
Again, I did not intentionally change anything that was already there. the only thing I did was add functionality by making it more responsive and providing audio feedback when things are not found, which is useful functionality.
I did not move things around as much you seem to think and it was absolutely not intentional to break things. Please try this version, it should work better except for the track and cuetrack trigger handling that seems to behave weird, although I have not been able to reproduce the problem.

View_Assist_custom_sentences/Play_Music_with_Music_Assistant/blueprint-playmusicwithmusicassistant.yaml
```diff
@@ -24,152 +24,188 @@ blueprint:
...
+                action: browser_mod.navigate
+              - target:
+                  entity_id: "{{ target_musicplayer_device }}"
+                data:
+                  media_type: artist
+                  enqueue: replace
+                  media_id: "{{ found_artist.artists.0.name }}"
+                action: music_assistant.play_media
+                enabled: true
+              - set_conversation_response: "\"Playing {{ found_artist.artists.0.name }} music now!\""
```
dinki, line 89: This needs to remain blank for now.  Folks using BrowserMod for their musicplayer_device will have issues.  BM does not handle playing annoucements over playing media files.   I plan to change this when I change the wiki to recommend HassMic for playback but for now this will need to be left as it was before to not break playback for others

View_Assist_custom_sentences/Play_Music_with_Music_Assistant/blueprint-playmusicwithmusicassistant.yaml
```diff
@@ -24,152 +24,188 @@ blueprint:
...
       - start [the] {playlist} playlist
     id: playlist
   - platform: conversation
     command:
       - play {song} by {artist}
-    id: playsong
+    id: track
   - platform: conversation
     command:
       - queue {song} by {artist}
```
dinki, line 45: My dumb STT recognizes this a 'Q' .   I have never used this before so I'm sure it's been broken for a long time.  Please change to:

- (queue | cue | Q) {song} by {artist}

View_Assist_custom_sentences/Play_Music_with_Music_Assistant/blueprint-playmusicwithmusicassistant.yaml
```diff
@@ -24,152 +24,188 @@ blueprint:
...
+description: "Play music using music assistant"
 trigger:
   - platform: conversation
     command:
       - play the artist {artist}
-      - play some {artist}
-    id: playartist
+      - play some {artist} [music]
+      - play [some] [music | songs] by {artist}
+    id: artist
```
dinki, line 34: I'm not sure why you changed all of the ids.   Change them back to original for playartist, playsong, cuesong

View_Assist_custom_sentences/Play_Music_with_Music_Assistant/blueprint-playmusicwithmusicassistant.yaml
dinki: All other responses will need to be changed back to empty.  Perhaps you can comment the current and add the empty one so that when the switch is made it will be easy to swap to the verbose one.

I am getting the word 'quote' at the end of the spoken text.  I think you are adding one too many quote characters

dinki:
Alright.  So I got my snapcast issues sorted.  Apologies as I think the last example I gave probably was working afterall.  I have tested all functions and all are working as expected.  I did make comments about some things that were changed unnecessarily (ids) and some prematurely (conversation responses).  

In summary, it looks like your PR adds a little more to the way the commands can be spoken, adds future conversation responses that will be commented for now and also gives the opportunity to add another STT option for the 'queue' command.  

If you can make those requested changes I think this one is ready to push in.  Thanks!

jeroenterheerdt:
Just left another comment for you above to ask you to do another test as I am unable to reproduce something @Flight-Lab mentioned here. https://github.com/dinki/View-Assist/pull/126#discussion_r1926220983

dinki:
Yeah, I did not test queue/cue/q due to the intent failing my STT .  I'll do a take control and add the 'q' wording and try it and see.  Apologies.  I did say all good but that has not been tested.  I'm running out of daylight so will give that a go tomorrow.

jeroenterheerdt:
did more updates. What I think is left is:
- figuring out that parameters are apparently not passed when queuing a song. I do not seem to have issues here.
- hopefully find a way to determine if users are on browsermod so we can give a nice response than just "done" which is the result of an empty response in my set up :(

dinki:
Looking at the BP in review and am thinking that any calls made from an audio_only device should not have the call to switch views.  I've done that like this on others:

```
          - if:
              - condition: template
                value_template: "{{ target_satellite_device_type != 'audio_only' }}"
            then:
              - data:
                  path: "{{ view }}"
                target:
                  device_id: "{{target_display_device}}"
                action: browser_mod.navigate   
```

By not including this the user could have the flow aborted.  To be clear, I'm not sure if that is 100% necessary as I use an HA VPE to play music and have not had a problem with it.  This is more a note to myself than anything.

jeroenterheerdt:
> Looking at the BP in review and am thinking that any calls made from an audio_only device should not have the call to switch views. I've done that like this on others:
> 
> ```
>           - if:
>               - condition: template
>                 value_template: "{{ target_satellite_device_type != 'audio_only' }}"
>             then:
>               - data:
>                   path: "{{ view }}"
>                 target:
>                   device_id: "{{target_display_device}}"
>                 action: browser_mod.navigate   
> ```
> 
> By not including this the user could have the flow aborted. To be clear, I'm not sure if that is 100% necessary as I use an HA VPE to play music and have not had a problem with it. This is more a note to myself than anything.

do you want me to include this or no?

jeroenterheerdt:
> @jeroenterheerdt I know you weren't able to reproduce it yourself, but I opened a PR in your branch that fixes the issue with the parameters not being passed. I'm unsure if this interferes with any future plans you have though.
> 
> [jeroenterheerdt#1](https://github.com/jeroenterheerdt/View-Assist/pull/1)

thank you @Flight-Lab this looks good to me.

So that leaves:
- hopefully find a way to determine if users are on browsermod so we can give a nice response than just a stupid "Done" which is the result of an empty response in my set up :(. @dinki any idea how we can detect the usage of browsermod?
- audio_only device comment here which I am not sure if we need to do anything with: https://github.com/dinki/View-Assist/pull/126#issuecomment-2613490122

dinki:
The spoken response that breaks BM speakers needs to be reverted to `conversation_response: ''` (from memory).  If you are getting a 'done' response from this then something else is wrong I am guessing as I know this is valid as I worked with Mike long ago to get it in.

The audio_only part is something I can do after this is pushed through.  It will be easier to do this after I think.

jeroenterheerdt:
I get that but I'd like to firs try to detect if a user is using BM. If they are we can return the empty response while keeping the more interesting response for those you don't use BM.

dinki:
Please modify as requested above.

jeroenterheerdt:
> Please modify as requested above.

No, you are just ignoring my suggestions. Instead of telling me to "do as I am told" how about actually answering my question?

dinki:
Here's your explanation:

> This needs to remain blank for now. Folks using BrowserMod for their musicplayer_device will have issues. BM does not handle playing annoucements over playing media files. I plan to change this when I change the wiki to recommend HassMic for playback but for now this will need to be left as it was before to not break playback for others

jeroenterheerdt:
> > ... I'd like to first try to detect if a user is using BM. If they are we can return the empty response while keeping the more interesting response for those you don't use BM.
> 
> It seems Home Assistant only tracks which integration "owns" an entity internally in the registry. In this case, it would be nice to have that information as an attribute attached to the entity. I suppose a python script could be written to retrieve that info from the entity registry, but that's another thing to ask the user to install.
> 
> An idea:
> 
> * Add a "select" selector to the device control blueprint.
> * User selects which media player they're using.
> * Media player type gets set as an attribute to be checked in this automation.

this is a way I think can work. Not doing stuff because catering for lowest denominator is not the right thing to do.

dinki:
Correct.  It is the announce that is not handled by BM.  I did not realize that Squeezelite was also problematic.  That changes a lot as deprecating BM will still leave the Squeezelite problem as active.  

What you suggest is a solution.  It is a bit complicated as this is an all-or-nothing option as if a user has a mix of players that do and don't work with announce, this switch turned on could cause the ones that don't work to be a problem.  Work could be done to include the speaker type in the control configuration per device but this is just adding too much complexity.  Those users will just have to decide the value in that response. 

So a select can be put in and users with all working will enjoy and those who have a mix or all BM can turn off and still function.  This is a good approach.

jeroenterheerdt:
Awesome, I'll try to finish up this PR first by putting the empty response back and then I will make a new PR to handle the spoken response + select in the blueprint. Thanks both!

jeroenterheerdt:
yeah, I got that. that's why I am finishing this up by putting it back to a state without a response so in a next PR we can put the "smart" response back and make it conditional on the control BP.

jeroenterheerdt:
ok, @dinki this should work now and not provide a response (`set_conversation_response: ""`).

Let me know if this is OK so we can move on the next step here :)

dinki:
> To be clear, what I suggested is not an all or nothing solution. I'm suggesting a select be put in the device control BP, not this BP. This BP would simply check for the attribute created by the device BP. This would make it work with a mix of media players and not an all or nothing thing. It would also allow for the future extension of responses during other automations, if so desired.

You are absolutely right.  I totally overlooked the obvious.  So an attribute would be set in the VA device based on this choice and that could be changed easily by the user if they choose to change their MP in the future to support.

@jeroenterheerdt Sounds like a good plan.  Would prefer to see that in a separate PR.  Thanks!

dinki:
@jeroenterheerdt We were typing at the same time.  I'll check this one out and hopefully push it through. 

Good stuff happening here guys.

dinki:
Okay, all tests passed so I will push this one through.  Thanks to you both.

Wait on me to update the device control for the additional attribute set in the device control blueprint.  I'll let you know when ready in Discord.  This will not be today but hopefully tomorrow.

jeroenterheerdt:
no worries, thanks for pushing this through!

PR #128, opened 2025-01-18
Adding another alternative clock, this time with movement

jeroenterheerdt:
Result of work done on Discord (https://discord.com/channels/1241796965344481440/1328102825938649169).
This is a version of the alternative clock but with a "screensaver-like" movement.

### DIFF
--- a/View Assist dashboard and views/views/clock/clockaltmovement.yaml
+++ b/View Assist dashboard and views/views/clock/clockaltmovement.yaml
@@ -1,0 +1,95 @@
+type: custom:button-card
+template:
+  - variable_template
+  - body_template
+variables:
+  clockaltcardversion: 1.0.0
+  var_background: >-
+    [[[ try {return
+    hass.states[variables.var_assistsat_entity].attributes.background} catch {
+    return  ""}]]]
+styles:
+  card:
+    - background: >-
+        [[[ return `center / cover no-repeat url(${variables.var_background})`
+        ]]]
+    - background-size: cover
+  grid:
+    - grid-template-areas: |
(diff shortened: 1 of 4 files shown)


jeroenterheerdt:
ideally in the future this should all be configurable without need to edit yamls. For example, people might want to show/hide dates/weather, disable/enable movement, etc.

dinki:
I'm thinking that this should probably be in Community Contributions views due to the need of editing YAML.  We can work towards getting this into core with the VA integration but for now I'd prefer not requiring folks editing YAML wherever possible.

Can you also update wiki to include how to configure?

jeroenterheerdt:
but we already have a clockalt there already? so shouldn't that then move as well?

dinki:
Clock alt was written by me so easily supported.  It does not take YAML configuration.  

I have some tips and tricks as well as automations that I plan to share as community contributions as they have dependency or very specific use and do not fit the core.

jeroenterheerdt:
I think once we have a panel all of these clock types (3 by now) are going to be just configurations on the clock view, so no need to edit yaml.

also, the clock with movement is just a view, there is no more yaml edits here than adding the original clock or the clockalt, or what am I missing?

dinki:
Is this view working?  If so, please submit as community contribution view as discussed.  If not, please close and resubmit when ready. I am trying to tidy up the outstanding PRs

jeroenterheerdt:
this view is functional for now. I moved it to community contributions as requested.

dinki:
Well done and well submitted.  Thanks!

PR #129, opened 2025-01-18
adding basic FAQ

jeroenterheerdt:
Adding a very basic FAQ in the docs. This list should grow as we get identify more FAQs


dinki:
Yep.  Good first step.  I'll have to think on this but I know a big one is folks using wrong BM name due to case.

PR #132, opened 2025-01-19
Functional spec

jeroenterheerdt:
Please review and leave comments :)


wiki/docs/developer-resources/functional-spec.md
```diff
@@ -0,0 +1,218 @@
...
+
+![Entering passcode](functional-spec-images/enter-passcode.png)
+
+The settings page contains tabs for general settings and view-specific settings. Each settings tab allows the user to make settings that are inherited by all satellites as well as to override settings per satellite. This gives the user maximum flexibility. All settings tabs use the same 'matrix' structure to visualize these settings.  By default the general settings tab is selected:
+
+![General settings](functional-spec-images/general-settings.png)
+
+By default the "global" setting value are inherited by all sattelites, but can be overriden per satellite by selecting the 'override' checkbox in the correct cell. That enables the UI element to change the default value, as the image above shows for the passcode setting for Satellite B. In the image above, all other global settings are inherited to both Sattelite A and B.
+
+Any setting changes made are applied immediately. The setting page can be closed using the same button as was used to open it. This button is available on all settings pages.
```
pantherale0, line 118: I think we should change the icon to an "X" when in the settings menu (make it look a bit clearer)

wiki/docs/developer-resources/functional-spec.md
```diff
@@ -0,0 +1,218 @@
...
+7. Specific weather entity to use (to override weather entity configured for all satellites)
+
+These settings are in the options flow because we don't feel like the user needs to change them that often after they are set up (principle ##5).
+
+## What the integration set up creates
+Succesfully completing at least one config flow of the integration creates the following items:
+- **a sensor** representing each satellite. Each satellite has its own sensor entity
+- **a panel** that hosts the views. There is only one panel and all satellites share the same panel but can have specific configurations for the views in the panel (see below).
+- **a group** hosting all the satellites. There is only one group of satellites.
+- **a timer** helper for each satellite. there is only one timer per satellite.
```
dinki, line 74: The need for a group can be eliminated as we can now list members of the VA integration instead.  I am thinking the need for multiple groups doesn't exist or would be more trouble than it is worth.  Current commercial assistants don't allow this

wiki/docs/developer-resources/functional-spec.md
```diff
@@ -0,0 +1,218 @@
...
+1. By default the `default view` is shown on the satellite's display.
+1. The user can activate other views by uttering access sentences. The view is shown on the satellite's device that the user is using.
+1. After `timeout` second the satellite in use switches back to the `default view`.
+1. Access sentences can be used while on any view, even while music is playing. If at all possible we would want to detect intents from the configured `assistant` and subscribe to handle those, so we do not need any automations.
+1. Something about link icons goes here, not sure yet what they are and how they fit in.
+
+## Development plan and priorities
+|Workitem #|Workitem|Priority|Pre requisites|Exit criteria|
+|---|---|---|---|---|
+|1|Integration and entity registration|0|None|User can select, load VA integration and add one or more satellites.|
```
pantherale0, line 197: When ready, let's split the integration piece up into seperate tasks using GH projects. We can do the same for the following work items too.

wiki/docs/developer-resources/functional-spec.md
```diff
@@ -0,0 +1,218 @@
...
+- **a group** hosting all the satellites. There is only one group of satellites.
+- **a timer** helper for each satellite. there is only one timer per satellite.
+- **services** that can be used to cause specific actions to happen on the satellites
+
+## Services
+The integration provides the following services:
+
+1. **Broadcast service**. This sends a message to all satellites which is shown on the intent view.
+1. **Set view service** (for all satellites or specific satellites). This switches the view on the specified satellites.
+1. **Set mode service**  (for all satellites or specific satellites). This sets the mode on the specified satellites.
```
dinki, line 82: What is special about setting mode and dnd?  Currently these are handled as setting an attribute and most BPs set this and other attributes with one call to set_state.py action.  Should we not just create a set_state.py equivalent (like we have currently in the integration) and use that?

wiki/docs/developer-resources/functional-spec.md
```diff
@@ -0,0 +1,218 @@
...
+|9|Add view|1|#3|User can add a view and invoke the view using view access sentences. Custom views can be added by pasting YAML code|
+|10|Access control|2|#3|User can enable / disable views from access by satellites and those changes are reflected|
+
+Not in the table above:
+- other view settings, similar to #7
+
+## Open questions
+Lots of open questions, but here are a few:
+1. Should we provide a way to customize access sentences for built-in views? Should this be a setting on the [view specific settings page](#view-specific-settings?
+1. Is it indeed possible to listen to intent from `assistant` and have the integration act on it? This is a research item which we might have to discuss with HA devs.
```
pantherale0, line 214: It would be good to see this integrated into core I think. I'll do some more research on the `assist_pipeline` component and see if I can get it to expose something on the event bus.

Event bus is important as it can contain more information about something than what a state change can do.

wiki/docs/developer-resources/functional-spec.md
```diff
@@ -0,0 +1,218 @@
...
+Add and cancel buttons are provided. If the user selects 'add view' the view is added to the tabstrip and if it is a built-in view its configuration settings are shown on that page.
+
+##### Access control settings
+The access control settings page allows the user to determine which views are accessible to which satellites. By default all views are available to all satellites. By disabling the correct checkbox a view can be made inaccessible to a satellite, which means that that satellite cannot access the view even if the correct view access sentence is uttered. The default view is always accessible by all satellites.
+
+![Access control](functional-spec-images/access-control.png)
+
+### Panel behavior
+1. By default the `default view` is shown on the satellite's display.
+1. The user can activate other views by uttering access sentences. The view is shown on the satellite's device that the user is using.
```
pantherale0, line 189: For future revisions we could implement something that allows touch control too, I know it's mentioned above but I'm thinking views that allow full touch control (could be an iframe for Music Assistant's UI for example). But not right now, let's get the basics working well first.

wiki/docs/developer-resources/functional-spec.md
```diff
@@ -0,0 +1,218 @@
...
+The settings page is where settings are made that have to do with how information is presented on the satellites. It is accessed from any view by a simple button press and can be protected with a passcode that needs to be entered to gain access. For example, here the user is accessing the settings from the clock view:
+
+![Accessing settings](functional-spec-images/accessing-settings.png)
+
+
+If a passcode is set up it needs to be entered correctly before the setting page can be accessed:
+
+![Entering passcode](functional-spec-images/enter-passcode.png)
+
+The settings page contains tabs for general settings and view-specific settings. Each settings tab allows the user to make settings that are inherited by all satellites as well as to override settings per satellite. This gives the user maximum flexibility. All settings tabs use the same 'matrix' structure to visualize these settings.  By default the general settings tab is selected:
```
pantherale0, line 112: We should agree how we store these settings, HA comes with some out of the box Python APIs we could use to pull in yaml / JSON from `.storage` / `config` for the integration side.

wiki/docs/developer-resources/functional-spec.md
```diff
@@ -0,0 +1,218 @@
...
+## Services
+The integration provides the following services:
+
+1. **Broadcast service**. This sends a message to all satellites which is shown on the intent view.
+1. **Set view service** (for all satellites or specific satellites). This switches the view on the specified satellites.
+1. **Set mode service**  (for all satellites or specific satellites). This sets the mode on the specified satellites.
+1. **Set do not disturb** service (for all satellites or specific satellites). This sets the specified sattelites to do not disturb or turns off do not disturb.
+
+## Panel
+This is the big one. The panel serves multiple purposes:
```
dinki, line 86: While I understand the desire, I think this is very ambitious.  We have users who use VA as a supplement to their current dashboard for their kiosk displays.  It seems what is being proposed would not accommodate.

Further, personally I do not want to configure anything on the VA devices themselves.  I loathe touching the screens and seldom even touch mine.  Can these settings be done within the HA web gui?

wiki/docs/developer-resources/functional-spec.md
```diff
@@ -0,0 +1,218 @@
...
+-  It's the dashboard satellites with a display will show. This will be configured automatically as part of the config flow for the satellite.
+-  The dashboard consists of multiple views. The following views are provided:
```
pantherale0, line 88: As a starting point, I think we should go for:

- Clock
- Alarm
- Intent
- Weather

Keep the deliverables small so we can get some users testing it out and providing feedback, then we know what the next requirements will be. Pretty much like what HA does with new features.

wiki/docs/developer-resources/functional-spec.md
```diff
@@ -0,0 +1,218 @@
...
+#### General settings
+The general settings tab provides settings that are not view specific. The following settings are provided. This is not an exhaustive list and will be expanded in the future. Also, this list is not necessary going to be shown in this order:
+
+|Setting name|Required yes/no|Type|Default value|Notes|
+|---|---|---|---|---|
+|**Passcode**|No|Whole Number|None|Passcode to access the settings page|
+|**Default view**|Yes|View|Clock|The home screen to show on the satellite|
+|**Timeout (s)**|Yes|20|Whole Number|Timeout in seconds after which the satellites switch back to its default view|
+|**Default mode**|Yes|Enum (normal, night)|Normal|Available options: normal, night|
+|**Show top bar**|Yes|Boolean|False|Show or hide bar at top of the screen to provide touch access to views, similar to top bar in HA dashboards that allows access to views|
```
dinki, line 135: This should stay within the third party display control integration

wiki/docs/developer-resources/functional-spec.md
pantherale0, line 74: I don't think this is possible, we will have to check with core team.

Timer isn't it's own platform in HA (unlike Sensor). But we might not need this if we can rely on events instead.

wiki/docs/developer-resources/functional-spec.md
```diff
@@ -0,0 +1,218 @@
...
+|**Default do not disturb state**|Yes|Boolean|False||
+|**Launch icons**|No|String|Empty string|Need to flesh this out!|
+|**Status icon size**|No|Whole number|?|Need to flesh this out!|
+|**Font style**|Yes|String|Roboto|Options: Roboto, neeed to provide dropdown of options|
```
dinki, line 139: This should remain free text to ensure maximum flexibility.  User can be guided to a website that lists available.

pantherale0 (commented):
A few initial comments - happy to discuss more

wiki/docs/developer-resources/functional-spec.md
```diff
@@ -0,0 +1,218 @@
...
+|**24-hour clock**|Yes|Boolean|Dependent on HA config||
+|**Assistant**|Yes|Assistant entity name|Default HA assistant||
+|**TTS**|Yes|HA entity name in `tts` domain|If only one TTS entity select that as default|The TTS entity to use for any system generated messages|
```
dinki, line 142: This is for TTS outside of Assist associated TTS?

wiki/docs/developer-resources/functional-spec.md
```diff
@@ -0,0 +1,218 @@
...
+|**Assist Prompt**|Yes|Enum(none, flashing bar, blur pop up)|Blur pop up|how the Assist prompt is visualized. Available options: none/flashing bar/blur pop up|
+|**Background image**|Yes|File name string|None|The background image|
+|**Rotate background**|Yes|Boolean|False|Indicates if backgrounds should be rotated. If true, further settings are required|
+|**Background rotation interval**|No|Whole number and time unit enum (second/minute/hour/day)|1 hour|Rotate background every XX time (optional, only applies if rotate background is on). This replaces the current blueprint XX minutes and XX hours settings. It allows the user to enter a value and select a time unit (seconds/minutes/hours/days).|
```
dinki, line 146: Agreed.  I don't like the current /5 formatting but I was either unable to work out something better or did not have time to find a workaround.  I think it was a limitation and was told it would not be an easy road.  Doing it here is great.

wiki/docs/developer-resources/functional-spec.md
pantherale0: I would suggest so yes - I had the same initial thought too, but realised things like broadcast isn't possible without defining the TTS entity.

Perhaps something we should investigate as an enhancement to core along side the `assist_pipeline`

wiki/docs/developer-resources/functional-spec.md
pantherale0: Looking at the UI, I would suggest these can be done from another browser.

I think custom views could accommodate this, but do remember that I think where we are talking about a view here, its not the same as a HA dashboard view.

The panel here would be something entirely custom in my view (typescript, html etc). A bit like what HACS / browser mod achieves.

We need to agree some terminology around this, I can see it getting confusing for us and users.

wiki/docs/developer-resources/functional-spec.md
dinki: This was meant for the group line not timer.  Apologies

dinki:
Closing for now.  We can reopen later.

PR #135, opened 2025-01-22
Update blueprint-devicefunctions.yaml

jeroenterheerdt:
found another issue while importing the blueprint after it was updated to include the music controls, fixing it


jeroenterheerdt:
still an issue. On import HA says:
```mapping values are not allowed here in "<unicode string>", line 204, column 98: ... authentication" style="word-wrap: break-word;"> ^``` and I am unable to spot the difference.

Putting it in yamllint gives an entirely different error which is not a part of the blueprint that I am familiar with and I doubt it is an actual error since seem to be an older change that put that in:
```Map keys must be unique at line 197, column 13```.

Any chance you can help?

jeroenterheerdt:
never mind, I will try to fix these issues but the current device functions BP that you merged this morning does not function :(

PR #134, opened 2025-01-21
Add music actions to device functions

jeroenterheerdt:
moving music control actions from musicassistant blueprint to device functions blueprint as requested in PR #126

PR #136, opened 2025-01-22
Fixing the device control blueprint that was broken after pr #134 was merged

jeroenterheerdt:
This works for me, can you test it before merging?


dinki:
Got it.  Thanks!

PR #137, opened 2025-01-22
More Music Controls

Flight-Lab:
Expanding music player controls from @jeroenterheerdt.

Add raise/lower as options in the music volume up and down commands.

Add mute/unmute commands.

Add set music volume command.


dinki:
Thanks for this for sure.  Does this also address:

https://github.com/dinki/View-Assist/pull/136

or does that need to be pushed too?

@jeroenterheerdt Please post here if you have tried these and they are working.  I will install later today as well but more testing the better.

Thanks to you both!

jeroenterheerdt:
#136 works, I have not tested this one and don't have time to test it at the moment :(

dinki:
Thanks for the great contribution!

PR #139, opened 2025-01-23
troubleshooting fixes

jeroenterheerdt:
As I was working on making sensors and groups work for the `intent` view, I needed the `troubleshooting` view to work...
As I was fixing `troubleshooting` view I discovered it has a lot of repetitive code and needs maintenance for new attributes added.

So I am also offering `troubleshooting-v2` that requires no maintenance as attributes are automatically updated and added. This version is maintenance free and much more elegant.

Note that both these troubleshooting view still rely on the assist group that is going away once the integration is updated, but either should be easily updated to just inspect that domain instead of the group.

Hope you find this helpful!

### DIFF
--- a/View Assist dashboard and views/views/troubleshooting/troubleshooting-v2.yaml
+++ b/View Assist dashboard and views/views/troubleshooting/troubleshooting-v2.yaml
@@ -1,0 +1,37 @@
+type: custom:button-card
+variables:
+  var_group: group.viewassist_satellites
+name: Troubleshooting (v2)
+show_icon: false
+show_state: false
+show_name: true
+custom_fields:
+  default_satellite: |
+    [[[ 
+      try {
+        if (!states[variables.var_group]) {
+          return 'Error: ${variables.var_group} not found';
+        }
+        let assistGroup = states[variables.var_group].attributes.entity_id;
+        if (!assistGroup) {
+          return 'Error: assist_group entity_id not found';
+        }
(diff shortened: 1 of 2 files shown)


dinki:
image

Thanks for working on this.  Something I've wanted to do as it is handy to have for new users.  I tried the modifications to the original and when I browse to it it is obscured by the the popup.  These lines  should be removed:

```
template:
  - variable_template
  - body_template
```

To explain a bit, I originally wrote this for folks who had issues getting things set up and were having trouble figuring out what was wrong.  This view is standalone as the template would not display properly due to misconfiguration by the user.  Having this view as a standalone helped figure out what was wrong.

Fast forward a bit and I've done some improvements to the dashboard that handle things better and errors are a bit easier to spot.  This view is still of value though.

I haven't looked at v2 yet.

jeroenterheerdt:
Interesting, I was more approaching this from the developer perspective but both audiences can be served with the same view.

dinki:
Hmm.. Odd for sure.  My brain is mush from a pretty long day at work.  Will look at it closer tomorrow.

jeroenterheerdt:
one thing I don't like about V2 is that the title is not shown (it just says: "Troubles..."). Not sure how to fix that.

But those are small issues.

dinki:
These should no longer be needed with the new integration but may be useful later.  Thanks for your efforts

PR #140, opened 2025-01-23
adding sensor support for intent view

jeroenterheerdt:
This adds support for sensors in the intent view.
Sensors are displayed using the built-in sensor card.

### DIFF
--- a/View_Assist_control_automations/blueprint-devicecontrol.yaml
+++ b/View_Assist_control_automations/blueprint-devicecontrol.yaml
@@ -478,7 +478,7 @@ action:
           {% set list =
           state_attr(intent,'intent_output')['response']['data']['success']
           | map(attribute='id') | list %} {% set prefix = ('light',
-          'switch', 'cover','input_boolean', 'fan', 'scene', 'group',
+          'switch', 'cover','input_boolean', 'fan', 'scene',
           'sensor') %} {% set data = namespace(filtered_list=[]) %} {% for
           item in list %}
             {%- if item.startswith(prefix) %}
--- a/View_Assist_control_automations/blueprint-devicecontrol.yaml
+++ b/View_Assist_control_automations/blueprint-devicecontrol.yaml
@@ -474,14 +474,29 @@ action:
     - service: python_script.set_state
       data:
         entity_id: '{{ satellite }}'
-        intent_entities: "{% set list = state_attr(intent,'intent_output')['response']['data']['success']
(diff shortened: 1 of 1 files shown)


dinki:
It's hard to track due to the crappy formatting from the original version that you have kindly cleaned up.  To be clear and to avoid me from having to look at all of those newly formatted lines, outside of that restructuring, the only other change is the addition of the sensor type right?

jeroenterheerdt:
you're right, let me highlight the changes I made for you.

1. I extended this line to include `sensor`:
```
{% set prefix = ('light',
          'switch', 'cover','input_boolean', 'fan', 'scene',
          'sensor') %}```

2. I added a IF statement in the `for` loop to test whether we are going to show a sensor or anything else. If a sensor, we return a sensor card, else we return what you originally had in here. The IF logic is actually reverse than what I just wrote here, but I hope it's clear:
```
{% if not e.startswith('sensor') %}
                {{ {'type': 'custom:button-card','entity':
                      e, 'tap_action':{'action':'toggle'}, 'double_tap_action':{'action':'more-info'}}
                }}, 
              {% elif e.startswith('sensor') %}
                {{ {'type': 'sensor','entity':
                      e, 'graph':'line', 'hours_to_show': 24}
                }},
              {% endif %}
              ```

jeroenterheerdt:
you're right, let me highlight the changes I made for you.

1. I extended this line to include `sensor`:
```
{% set prefix = ('light',
          'switch', 'cover','input_boolean', 'fan', 'scene',
          'sensor') %}```

2. I added a IF statement to test whether we are going to show a sensor or anything else. If a sensor, we return a sensor card, else we return what you originally had in here. The IF logic is actually reverse than what I just wrote here, but I hope it's clear:
```
{% if not e.startswith('sensor') %}
                {{ {'type': 'custom:button-card','entity':
                      e, 'tap_action':{'action':'toggle'}, 'double_tap_action':{'action':'more-info'}}
                }}, 
              {% elif e.startswith('sensor') %}
                {{ {'type': 'sensor','entity':
                      e, 'graph':'line', 'hours_to_show': 24}
                }},
              {% endif %}
              ```

dinki:
This one was hand edited to make my life easier.  Thanks!

https://github.com/dinki/View-Assist/pull/148

jeroenterheerdt:
Did you figure out how to notify users of changes in views?  I am aware now but not everyone is. Could you consider creating a release?

PR #149, opened 2025-02-12
Add mode change according to musicplayer_device status in blueprint-deviceco…

relust:
…ntrol.yaml

I found that when the music view is changed, it doesn't go back to the home/clock view after the song is finished playing or if the speaker is turned off. For the `musicplayer_device` I use a different media player than that of the ViewAssist satellite, i.e. google speakers. I think it would be useful to switch the view from the viewassist satellite according to the status of the `musicplayer_device` speaker. Even if I start the playback on the speaker by other methods than the viewassist automations, I think it would be useful for the satellite tablet to see what is playing on the speaker. I've added two triggers `musicplayer_playing` and `musicplayer_not_playing` and two actions to fire the `python_script.set_state` and update the `music mode` or `normal mode`. Please look into whether this functionality is implemented somewhere and give me an answer.

### DIFF
--- a/View_Assist_control_automations/blueprint-devicecontrol.yaml
+++ b/View_Assist_control_automations/blueprint-devicecontrol.yaml
@@ -329,6 +329,24 @@ trigger:
 - platform: time_pattern
   hours: !input rotate_hour
   id: rotateimage
+- alias: Musicplayer playing
+  id: musicplayer_playing
+  value_template: >
+    {% set ent = state_attr(trigger_satellite, 'musicplayer_device') %} {{ ent is not none and is_state(ent, 'playing') }}
+  trigger: template
+  for:
+    hours: 0
+    minutes: 0
+    seconds: 30
+- alias: Musicplayer not playing
+  id: musicplayer_not_playing
+  value_template: >
+    {% set ent = state_attr(trigger_satellite, 'musicplayer_device') %} {{ ent is not none and not is_state(ent, 'playing') }}
+  trigger: template
+  for:
+    hours: 0
+    minutes: 0
+    seconds: 30
 variables:
   satellite: !input satellite
   group_entity: !input group_entity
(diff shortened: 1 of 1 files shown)


dinki:
What happens if the musicplayer and mediaplayer are the same device?  If the mediaplayer starts playing a TTS or maybe the alarm sound for a timer, would this be triggered?  I think it would.

relust:
I haven't tried this yet. Probably a secondary sound or song is interrupting the first one. I set that the mode change is done only if 30 seconds of playing or not playing are detected so that the mode is not changed for a simple announcement or a short pause of media player.

dinki:
I don't think this is going to work.  I could see this having false triggers.  This can be a custom automation but I don't think it will fit into the core as it stands.

relust:
- My problem is not with auto ON music view, but rather with auto OFF, because after a song is finished playing or the speaker is turned off, it does not return to home view/clock view but remains music view. I couldn't get it to change to "normal mode" when the playback is finished from the same automation where I start the music playback because the response to the conversation, for example "Music is playing in the office", is given not when the command is given but when the automation finishes executing. For this, a new automation had to be done, and that's why I thought that we could also include this in the control display automation so that no additional automation had to be done for each satellite.
- For auto On, I set music mode to be activated only if the media player is in the playing state for more than 30 seconds. I don't think any tts message should be longer than 30 seconds, but we could set more time instead of 30 seconds if needed.

relust:
@Flight-Lab I looked at your Extended device controls blueprint and it looks really good. I will install it myself. I really think it is a solution to the problem I raised. I'm going to try to include the automation I have for custom responses in this blueprint of yours, and if it goes well, I'll submit a PR to this blueprint.

relust:
If we can use the Extended device controls blueprint, I don't think this PR of mine is needed anymore, but I still feel that finding a way to change the mode to normal mode when it finishes playing shouldn't be solved separately.

PR #152, opened 2025-02-25
Extended device control update STT input description and fix auto music mode on/off

Flight-Lab:
Input description to include hassmic example.

Fix the broken template in auto music mode on/off.

Music mode on/off now runs navigate and set state actions in parallel.

Bump version to 1.1.1


dinki:
Thanks for your continued support!

PR #156, opened 2025-04-10
Fix popup UI effects for pre-wyoming HassMic devices

Flight-Lab:
This PR fixes issues with the blur popup and other UI effects not working properly with older versions of HassMic.
Problem
The template logic in dashboard.yaml only checked for specific states with Stream Assist and HA Voice Satellite devices, but lacked proper detection for HassMic states.
Solution
Updated the var_assisting template to properly detect HassMic active states:

- Added detection for "stt-listening", "intent-processing", and "tts-speaking" states.
- Maintained compatibility with other microphone types.
- Restored functionality that existed in older versions.


dinki:
Thanks for this.  I did just remove this from code the other day not thinking I was breaking stuff for folks who may be using the old version.  At some point this will go away when HM Wyoming is more stable.

PR #167, opened 2025-04-19
Revised german translation in blueprint-howstheweather.yaml

datenzar:


dinki:
Thanks for making VA better!

PR #168, opened 2025-04-19
Translation mechanism for date queries

datenzar:
- added translation maps for weekdays and months in en and de
- switched language input into selector


dinki:
Will look at the things mentioned earlier.  If you have time, can you look at converting the selector to use the language name (eg French) but provide the same value used (eg fr).  Here's an example I used previously:

```
          selector:
            select:
              custom_value: true
              options:
                - label: Small
                  value: 6vw
                - label: Medium
                  value: 7vw  
                - label: Large
                  value: 8vw
```

What do you thing?

datenzar:
> - label: Small
>                   value: 6vw
>                 - label: Medium
>                   value: 7vw

Not sure that I can follow. Do you mean the selector drop-down for languages? I used the build-in language selector to support a nice dropdown like this
grafik

dinki:
I'm going to push this one through and make a slight mod to it.  I will also close the other PR that contains the jinja with dependency.  Thanks again!

PR #173, opened 2025-04-21
German Translation to Device Functions

datenzar:


dinki:
Thank you.  This BP is a work in progress so will need additional help once I finish the others.

Note that these will need to remain in english as these modes are hardwired English in the view names and the modes.
```
## Modes

- en: `"['normal', 'night', 'hold', 'cycle']"`
- de: `['normal', 'nacht', 'halten', 'zyklus']`

## Views

- en: `"['clock', 'weather', 'info', 'infopic']"`
- de: `['Uhr', 'Wetter', 'Info', 'Infobild']`
```
We can, however, create a dictionary versus a list so that the VA specific values are sent to the VA even when using the language specific version.  Something like:
```
- en: '{"normal": "normal", "night": "night", "hold": "hold", "cycle": "cycle"}'
- de: '{"normal": "normal", "nacht": "night", "halten": "hold", "zyklus": "cycle"}'
```

datenzar:
That sounds like a good approach. I have removed the modes and views translation from the README for now

dinki:
This is great.  Thank you for your continued contributions!

PR #174, opened 2025-04-21
Translation of "What Time Is It" blueprint using easy_time.jinja macros

datenzar:


dinki:
`{% from 'easy_time.jinja' import weekday %}`

Does the user need to install something for this to be available?

dinki:
I absolutely hate dependencies.    Perhaps if this is mostly static and we can figure out an easy way to install/update then can give it a thought but would prefer another method unless it is WAY too complicated.

I haven't looked really close at this but is this only for swap from 12 hour to 24 hour and vice versa?  If so, I have an easy fix for that.

dinki:
I've made the change to the time variable using the VA attribute as the condition.  I have not tried it but I think it should work.  I'll try it tomorrow but if you like you can give it a go and see if it works as you expect it.

datenzar:
> I absolutely hate dependencies. Perhaps if this is mostly static and we can figure out an easy way to install/update then can give it a thought but would prefer another method unless it is WAY too complicated.
> 
> I haven't looked really close at this but is this only for swap from 12 hour to 24 hour and vice versa? If so, I have an easy fix for that.

I understand your worries. I will see what's required and if the 12/24h approach is helpful.

dinki:
Closing this one in favor for the other PR just merged.  Thanks.

PR #177, opened 2025-04-26
Sports score italian translations

nu00:
Added the Italian translation (not tested yet). Keep in mind that in Italian, some team names are masculine and others are feminine, which affects the choice of articles. I couldn't find a reliable way to handle this automatically, so I avoided using articles where possible.


nu00:
If anyone could test it, that would be appreciated! I'm not really into sports myself, so I'm not even sure what to ask. :)

dinki:
Yeah, understood and thought those articles (??) could be problematic.

dinki:
This looks good to me!  Thanks again!  Merging.

PR #176, opened 2025-04-23
Feat list translation and list view scrolling

rummens:
## List Blueprint

- Added translation for German, Spanish and French.
- Added new option to only speak out a certain number of items on the shopping list
- Added new option to overwrite the name of the shopping list (Mine has a very long name as I use an integration with Thermomix and cannot change the name of the list)
- Added proposal for new README structure and updated wiki accordingly.

PR #179, opened 2025-04-27
Feat list translation and list view scrolling (right branch)

rummens:
Corrected branch of #176

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,7 @@
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+
+# IDEs
+.vscode
+.idea
--- a/View Assist dashboard and views/views/list/list-nocheckbox.yaml
+++ b/View Assist dashboard and views/views/list/list-nocheckbox.yaml
@@ -27,7 +27,14 @@ custom_fields:
               box-shadow: none;
               border: none;
               position: absolute;
-              top: 15vh;
+              top: 15vh; /* Play with this value to adjust top spacing */
+              height: 80vh; /* Limit the vertical size to allow scrolling */
+              /* flex-direction: column-reverse;  <-- Enable this to auto scroll to the end of the list */
+              overflow-y: auto;
+              padding: 1rem;
+              display: flex;
+              flex-wrap: wrap;
+              align-content: flex-start;
             }
             ha-check-list-item {
               color: white;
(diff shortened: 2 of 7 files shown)


dinki:
This is great!  Thank you again.

If I can push my luck, I have a couple things to note.  If you can make these changes it would be great but if not I can certainly do them.

Can you move the [set_conversation_response](https://github.com/dinki/View-Assist/blob/3b7b356a2e8cd3ace7b7d359e45bd69e14e47e3c/View_Assist_custom_sentences/List_Management/blueprint-listmanagement.yaml#L152-L154) to a conversation_response variable and then use that for the conversation response and also set the last_said variable.  [Here's an example}(https://github.com/dinki/View-Assist/blob/viewassist-multilingual/View_Assist_custom_sentences/Locate_a_Person/blueprint-locateaperson.yaml#L90-L104)

Can you move the language select to use the pulldown with 'friendly name' versus the two letter language code [like this](https://github.com/dinki/View-Assist/blob/viewassist-multilingual/View_Assist_custom_sentences/Locate_a_Person/blueprint-locateaperson.yaml#L6-L15)

Lastly, I am seeing where I can add items that are already on the list to the list.  Example 'eggs' and 'Eggs'.  I thought I was handling case correctly but maybe not.  If you want to check this that would be great too.

Thanks again for your support on this effort.  I am trying to balance my time on all of this and your help is most appreciated and not taken for granted.  Again, just let me know if you do not have time or don't care to work on these items and I'll polish them.

rummens:
Hey there,

sorry for the long silence, its been a long week :)

I pushed all changes and even added one more new item, which is some documentation to reverse the list sequence. In my testing I was always annoyed by the fact that the oldest items are on top. So I added some steps to help with that, its not perfect since this is a limitation by the to-do card but the current solution gets at least close.

On the comparison topic: can you give me some example sentences that dont work for you? I tried with a few items in German and English but can't replicate the behaviour. It correctly identifies the item already being on the list and not adding it again.

image

dinki:
Looks really good.  I removed everything from my list and started from zero and it appears to not be adding duplicates.  Not sure what I had going on but this seems to have fixed it.  Thanks for your efforts.  I'll push this one through!

One side note, a user provided a better method for setting the input variable than the one I suggested using a built in:

```
    language:
      name: Language
      description: The desired spoken language
      default: en
      selector:
        language:
          languages: [de, en, es, fr, it, nl]
```

I'll make that little change after merging.  Thanks again!

PR #180, opened 2025-04-27
Feat/sport translation German, Spanish and French

rummens:
Based on the work done, I added the German, Spanish and French translations. I also moved out the trigger sentences to a var and made proposals for people to overwrite them.

Sorry for the small commit mess, I had to rebase my branch after figuring out I based it again on main.

@nu00 Can you take a look at a small modification I did on the Italian version? I think the code merges a sentence and the outcome. So e.g. `{team_name} played {opponent_name} and` + `won`. Now I don't speak a word of Italian but I figured its missing and "and" there for the final sentence to make sense. But I might be wrong?


nu00:
Yes, you're right, I misinterpreted the sentence. I'll take a look.

nu00:
Your version is correct. However, there is still the issue that some team names are feminine and others are masculine. In English it's not a problem since "The" is always the same, but in Italian it can be "La {team_name_femminine}" or "Il {team_name_masculine}". There is no general rule for this, so the only reliable solution might be to create a predefined map linking each team to the correct article. For now, I have omitted the article to avoid awkward or cacophonous phrasing.

dinki:
Looks good!  Thank you!

PR #181, opened 2025-05-04
Added italian support for person location

nu00:
Still needs to be tested, also fixed the text response in case a person is not defined to match the voice response.

### DIFF
--- a/View Assist dashboard and views/views/locate/locate.yaml
+++ b/View Assist dashboard and views/views/locate/locate.yaml
@@ -95,27 +95,10 @@ custom_fields:
             return `${var_location_text}`}
           catch { return ""}]]]           
         updated: |-
-          [[[ try {
-            var var_tracker = hass.states[variables.var_assistsat_entity].attributes.locate_data.person;                  
-            var var_last_changed = new Date(hass.states[var_tracker].last_changed);
-            const current_date = new Date();
-            const diff = current_date - var_last_changed;
-            const minutes = Math.floor(diff / (1000 * 60));
-            const hours = Math.floor(diff / (1000 * 60 * 60));
-            const days = Math.floor(diff / (1000 * 60 * 60 * 24));
-            const rtf = new Intl.RelativeTimeFormat('en', { numeric: 'auto' });
-            let last_changed;
-            if (minutes < 1) {
-              last_changed = "now";          
-            } else if (minutes < 60) {                    
-              last_changed = rtf.format(-minutes, 'minute');  // e.g., "5 minutes ago"
-            } else if (hours < 24) {
-              last_changed = rtf.format(-hours, 'hour');  // e.g., "2 hours ago"
-            } else {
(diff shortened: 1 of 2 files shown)


nu00:
It would be nice to translate "home" into the native language. Can I go on?

dinki:
Looks great!  See you also corrected a typo I missed.  Thanks for this!

Where is the 'home' issue?

nu00:
When a user is at home, their state is displayed as 'home'. However, in Italian, 'home' translates to 'casa'. While the state for other locations typically reflects the name of the corresponding zone, the home zone is consistently shown in English. It would be advisable to localize the label 'home' according to the user's language settings.

Furthermore, the Location view should also offer support for multiple languages.

dinki:
Ah for the view.  I'm open to this for sure.

nu00:
This applies both to the view and to the blueprint. If I ask 'Where is John?' and John is at home, it will always return 'home' in English, regardless of the language. We could use something like `{{ state_translated("person.name") }}` to provide a localized version.

dinki:
This is a HA problem with the zone though right?  Even in your target language HA will show 'home' or is that incorrect?

nu00:
Yes, that's correct. If you rely on the state of the person entity, it will always return home, not_home, or the exact name of the zone the person is in. The label home is not automatically translated, even if the user interface is in another language.

I was thinking of using something like this (tested and working):

```
variables:
  person_source: "{{ 'person.' + persons_name[trigger.slots.person] }}"
  tracker_source: "{{ state_attr(person_source, 'source') }}"
  geocoded_source: >-
    {% set geocoded = 'sensor.' + tracker_source.split('.', 1)[1] + '_geocoded_location' %} {{ geocoded }}
  location_source: |-
    {% if states(person_source) != 'not_home' %}
      {{ state_translated(person_source)|capitalize }}
    {% elif not states(geocoded_source) in ['unknown'] %}
      {{ states(geocoded_source) }}
    {% else %}
      {{ 'Unknown Location' }}
    {% endif %}
  conversation_response: |-
    {% if states(person_source) == 'not_home' %}
      {{ translations[language]['responses']['person_away'].replace("{person}", persons_name[trigger.slots.person]) }}
    {% else %}
      {{ translations[language]['responses']['person_location']
        .replace("{person}", persons_name[trigger.slots.person])
        .replace("{location}", state_translated(person_source)) }}
    {% endif %}
```

dinki:
I think that would work out great!

nu00:
Ok, done! I've tried it and it seems to work for me. Maybe the version number needs to be changed. If anyone wants to try it, that would be great!

nu00:
As we discussed on Discord, I’ve made a few changes to the view to accept the last change string from the Blueprint, translated into the language selected in the Blueprint.
Now the Blueprint calculates the last update and builds the already translated string, passing it to the view via set_state

dinki:
Fantastic!!  Thank you!!

PR #188, opened 2025-05-19
Update m5stack-atom-echo.yaml fix for framework version

tohurtv:
Added quick fix to pin the framework version due to the lighting control used in this config and the updated features for new framework version don't quite work yet but pinning the version allows the config to work on the current version of ESPHome


dinki:
Thanks for this update!

PR #189, opened 2025-05-29
Update m5stack-atom-echo.yaml and documentation.

tohurtv:
I have updated the m5stack-atom-echo.yaml  to completely follow the latest ESPHome conventions and cleaned up the documentation.


dinki:
Thanks for the updates and making this available for all!

PR #194, opened 2025-06-20
Camera View & Blueprint Enhancement: Multi Camera Support, URL Based Navigation, Configurable View Timeout, Expanded Voice Controls

Flight-Lab:
## Overview 
This PR introduces a comprehensive overhaul of the View Assist camera view functionality, transforming it from a single camera attribute based system to a modern, responsive multi camera interface with URL based navigation and configurable timeout management. Uses native picture entity card to reduce dependencies. Blueprint allows for setting multiple aliases for each camera entity, as well as the ability to show a single specific camera or a configured list of cameras with custom ordering.

### Multi-Camera Grid Display
- **Responsive Layout**: Automatically adjusts to 1, 2, or 3 columns based on screen width
- **Camera Discovery**: Automatically detects and displays all available cameras
- **Configurable Camera Lists**: Support for showing specific configured cameras with custom ordering
- **Grid Navigation**: Touch friendly grid interface for easy camera selection
- **Smart Auto-Display**: Automatically shows single camera when only one is available

### Enhanced Voice Commands
- **"Show all cameras"**: Display responsive grid of all cameras specified in blueprint configuration
- **Multiple Camera Names**: Each camera entity can have multiple voice command aliases
- **Intelligent Fallback**: Configurable behavior when camera names aren't found
  - Show configured cameras (default)
  - Speak available camera names
  - Do nothing

### URL-Based Navigation System
- **Stateless Design**: Replaces persistent attributes with URL parameters
- **Direct Camera Access**: `?camera=entity_id` for specific cameras
- **All Cameras Grid**: `?show=all` for complete multi-camera display
- **Configured Cameras**: `?show=configured&cameras=list` for custom camera subsets
- **Timeout Integration**: `?timeout=seconds` for hold mode duration

### Configurable Hold Mode Timeout
- **Frontend Timer System**: Reliable JavaScript-based timeout management
- **Flexible Duration**: 0-3600 seconds (0 = infinite hold)
- **Auto-Cleanup**: Prevents memory leaks on navigation
- **Page-Specific**: Avoids duplicate timers across views

### Frontend Enhancements
- **Native Components**: Removes dependency on `advanced-camera-card`
- **Picture Entity Cards**: Uses Home Assistant's native `picture-entity` cards
- **Responsive Design**: Adaptive layouts for camera grid
- **Memory Management**: Proper cleanup of timeouts and event handlers
- **Configurable Camera Ordering**: Support for custom camera list display order

### Architecture Changes
- **Stateless Design**: URL parameters replace persistent entity attributes
- **Modular Variables**: Clean separation of concerns in template variables
- **Enhanced Camera Logic**: Support for both automatic discovery and configured lists
- **Error Handling**: Robust error handling for camera availability
- **Performance**: Efficient camera discovery and rendering

### Blueprint Configuration Format
```yaml
# OLD FORMAT
cameras:
  front: camera.front_camera
  doorbell: camera.doorbell_cam

# NEW FORMAT
cameras:
  camera.driveway:
    - driveway
    - street
  camera.doorbell_cam:
    - doorbell
    - door
    - entrance
```

### Removed Features
- **Hold Mode Input**: Now automatically managed by the view
- **Camera Attribute**: Replaced with URL parameter navigation
- **Advanced Camera Card**: Now uses native Home Assistant components

### Navigation Flow
1. **Voice Command** → Specific camera or configured cameras grid view
2. **Grid Selection** → Touch camera thumbnail to view full screen
3. **Back Button** → Return to grid from individual camera
4. **Auto Timeout** → Automatically revert to default mode from hold mode (configurable)

### Enhanced Camera Management
- **Custom Camera Lists**: Show only the cameras you want in your preferred order


dinki:
@Flight-Lab Also, apologies, but I did make some updates to main and just pushed the PR to sync with dev.  You'll want to update your branch to avoid pushing changes to those already updated files.  I'm still real shaky with all of this.

Flight-Lab:
> @Flight-Lab Also, apologies, but I did make some updates to main and just pushed the PR to sync with dev. You'll want to update your branch to avoid pushing changes to those already updated files. I'm still real shaky with all of this.

Already done, I appreciate the heads up none the less.

dinki:
> > @Flight-Lab Also, apologies, but I did make some updates to main and just pushed the PR to sync with dev. You'll want to update your branch to avoid pushing changes to those already updated files. I'm still real shaky with all of this.
> 
> Already done, I appreciate the heads up none the less.

Still seeing 37 files changed on this PR though.  I'm guessing you only really changed the view and BP for camera with your latest commit right?

dinki:
I totally missed your big response yesterday.  Apologies for the slow response and perhaps looking like I was ignoring them.  I did not see them until now.

> Rather than relying on individual blueprints to explicitly handle title cleanup, which only covers voice-triggered scenarios and leaves manual navigation unaddressed, this feels like something that should be built into the navigation system itself. The `view_assist.navigate` service could automatically clear contextual attributes like `title`, `message`, and `list` on navigation, ensuring clean state transitions regardless of how navigation occurs (voice, manual touch, timeout revert, etc.).

You are right.  I do very little in regard to touch on these displays so VA is not handling this 'correctly' for those who use touch.  I think the best approach would be to clear these on timeout revert.  I think that could be done much easier than with navigate as that service call is made after attributes are set in the BPs.  Thoughts?

> Great point, we can also use this to set the order they display in the 'all cameras' view based on the order they're added in the blueprint.

The listing and ordering would be really cool.  By default you can just show all cameras but use this custom dictionary if the user does provide it.  It would also, I think, make this potentially extendable to using the blueprint for groupings.  Like 'show all backyard' cameras and it would show those in the all group for that automation built from the master BP.

> I noticed a delay in loading with the Advanced Camera Card as well, but it does appear to take a second longer with the Picture Entity Card (this was tested side by side in a non View Assist test dashboard consisting of just these two cards displaying the same camera feed). Another drawback of the Picture Entity Card is that it doesn't enable audio playback of the camera feed without having to click the camera feed to open the more info dialog. This shows an even smaller live feed and covers up the status icons in the upper right corner of the view. For this reason, I plan to create an alternate version of this view using Advanced Camera Card for my personal use and submit that as a community contribution, or follow what was done with the clockalt.yaml and put it in the same folder as the main camera.yaml, depending on which you'd prefer.

I think I prefer the Advanced Camera Card but do not like the dependency.  Perhaps using this camera view as the default but also providing the ACC version as a supplemental would work great.  If it is possible to use the same BP and user just change the destination view it would make things pretty tidy.  The wiki would list the required dependency but the default would work without any additional work.

> Correct, the only files I changed were the view and BP for camera. This branch is synced with 'dinki/View-Assist:main' and all the other changed files are your commits.

I think everything will merge fine even with the duplicates so believe we are all good.  Thanks for your continued support and sharing!

Flight-Lab:
> I totally missed your big response yesterday. Apologies for the slow response and perhaps looking like I was ignoring them. I did not see them until now.

No worries, I never once considered such. It is always a pleasure working on this project, and understand we all lead a life outside of it.

> I think the best approach would be to clear these on timeout revert. I think that could be done much easier than with navigate as that service call is made after attributes are set in the BPs. Thoughts?

If it's only cleared on timeout revert, the issue will still exist for navigation calls from other sources and continue to require `set_state` calls to clear these attributes in each blueprint. My thoughts were to just change the order of the calls in the blueprints and have the attributes get set after the navigation call was made.

Maybe we should make this an issue to work on separately, and I'll just add your suggestion to "...do a set_state for title: '' attribute", in order to get this PR merged sooner. That's up to you, of course.

> The listing and ordering would be really cool. By default you can just show all cameras but use this custom dictionary if the user does provide it. It would also, I think, make this potentially extendable to using the blueprint for groupings. Like 'show all backyard' cameras and it would show those in the all group for that automation built from the master BP.

I have a working branch with the listing that I'm going to merge in later this evening once I've tested it some more. The grouping is a really nice idea, I'll likely submit that in a separate PR though.

>  Perhaps using this camera view as the default but also providing the ACC version as a supplemental would work great. If it is possible to use the same BP and user just change the destination view it would make things pretty tidy. The wiki would list the required dependency but the default would work without any additional work.

Sounds like a great idea. Shouldn't need any change to the BP to implement exactly how you describe.

I appreciate your appreciation! More than happy to share what I've done for myself to save others the time.

Flight-Lab:
> @dinki if you can confirm that the issue with the stale title persisting has been fixed, this PR should be fully caught up on your requests (other than camera grouping) and ready to be merged.

I have tested this and can confirm the stale title no longer persists when navigating by voice with the blueprint. As expected, and as before, the stale title persists through all other means of navigation. This issue is described in further detail in https://github.com/dinki/view_assist_integration/issues/144

PR #196, opened 2025-07-01
Update clockaltwithmovement.yaml

jeroenterheerdt:
Updating the alt clock contribution so it works with the View-Assist integration (it took me a while to take the plunge).


jeroenterheerdt:
also fixing a small typo in the wiki where the title on [this page](https://dinki.github.io/View-Assist/docs/community-contributions/cc-sentences/conversions) is incorrect. It should be Conversions.

PR #208, opened 2025-08-12
feat: add camera status icon template

Flight-Lab:
New predefined icon navigates to camera view.

Uses mdi:cctv


dinki:
What defines the camera that is shown?  Shows last used?

Flight-Lab:
It shows the all cameras view, displaying all cameras configured in the blueprint, the same as requesting to see all cameras by voice.

If the blueprint has not been setup yet, it shows all existing camera entities. This is mostly a fallback for new users that may be skipping steps.

The logic for this already exists in the view.

PR #211, opened 2025-08-13
Reorder navigation and title setting in blueprints

Flight-Lab:
**PR is a draft as testing is still needed. Testing should show that the titles display as intended.**

Reorders `navigate` and `set_state` calls in blueprints to set titles after navigation instead of before. This ensures titles survive the navigation auto-clear feature added in the integration.
Addresses issue https://github.com/dinki/view_assist_integration/issues/144
Requires PR https://github.com/dinki/view_assist_integration/pull/224 that implements auto-clear of title attribute during navigation. These changes work together:
- Integration: Auto-clears `title` during `async_browser_navigate()`
- Blueprints: Set titles AFTER navigation to prevent clearing

## Changes
**Updated blueprints to set titles AFTER navigation:**
- Search Wikipedia: Reorder for both infopic and info view sequences  
- List Management: Reorder in add, remove, and show list sequences
- Broadcast: Reorder in repeat loop for announcement title

**Removed explicit title clearing workarounds:**
- Show Webpage: Remove `title: ""` manual clearing
- Spell a Word: Remove `title: ""` manual clearing  
- View Camera: Remove `title: ""` from all camera sequences

## New Pattern
Blueprints that need contextual titles now follow the pattern:
1. Navigate first (clears stale context)
2. Set title/content after (survives navigation)


Flight-Lab:
I think it just happens so quickly / the view takes a moment to load that it isn't even noticed. If you are worried about that we can split the set state calls in two, allowing the title to be set after 
I think most of the "other stuff" is view specific (like list or any of the display_timer_xxx) or likely explicitly set in the view or dashboard (like message and message font size). One that may be a concern is image in the search Wikipedia BP, where I could see an image holding over from one search into another that does not contain its own image to use (if this isn't already accounted for in the BP). Of course, we can include any of these attributes to be cleared along with the title if you think there will be issues.

PR https://github.com/dinki/view_assist_integration/pull/163 has been tested to perform as expected, but as mentioned above, most of these BPs still need to be tested if you'd like to have a go.

dinki:
Yeah I guess it would not kill us to break out just the title so that it is set after and keep the other stuff where it is now.  I may be overthinking it but I really want to ensure we don't see stale data on low power machines. 

> PR https://github.com/dinki/view_assist_integration/pull/163 has been tested to perform as expected, but as mentioned above, most of these BPs still need to be tested if you'd like to have a go.

I'm going to have to carve out some time to get this tested.  Hopefully it won't take too much time. We are a madhouse at work and weekends pretty nuts too but will try to get this going.

PR #236, opened 2025-09-17
Update for the Xiaomi Smart Clock documentation

lflondonol92:
Adding an optional step to reduce the boot time in half


dinki:
Thanks for the update!

PR #237, opened 2025-09-17
Spanish translations for weekdays and months for What time is it

lflondonol92:
New dictionary for the days of the week and months in spanish


dinki:
Thanks for the contribution!

PR #257, opened 2025-11-10
Add template validation for status icons

Flight-Lab:
## Description
Prevents button-card errors when users add custom status icons without corresponding templates in dashboard.yaml. Instead of cryptic errors, displays a user-friendly orange warning indicator showing which template is missing.

## Changes
- Query lovelace config to detect available button-card templates
- Validate static icon templates before rendering
- Show orange alert icon with dashed border for missing templates
- Dynamic icons (view:, entity:, service:) unaffected - always use core templates
- Graceful fallback if template detection unavailable


dinki:
Thanks for this!!

PR #267, opened 2025-11-20
Fix(camera-bp): Include vaca device type in camera list restoration on startup

Flight-Lab:
## Description
Fixes an issue where VA devices with the `type` as `vaca`, would lose their camera list configuration after Home Assistant restarts or integration reloads.

## Problem
The camera view blueprint's startup restoration sequence only filtered for `view_audio` device types, excluding `vaca` device types. This caused VACA users who manually navigated to the camera view before using a voice command to see all cameras displayed rather than only those in their configured list, since the `camera_list` attribute was never restored on startup for the `vaca` `type` devices.

## Solution
Expanded the `view_assist_entities()` filter on the startup trigger to include both `view_audio` and `vaca` device types, ensuring all View Assist devices, (except the `audio_only` type), receive their configured camera lists during Home Assistant startup.

### DIFF
--- a/View_Assist_custom_sentences/View_Camera/blueprint-viewcamera.yaml
+++ b/View_Assist_custom_sentences/View_Camera/blueprint-viewcamera.yaml
@@ -205,5 +205,3 @@ action:
                       last_said: "Camera '{{ requested_camera }}' not found."
             default: []
 mode: single
-
-


Flight-Lab:
Closed in favor of https://github.com/dinki/View-Assist/pull/268 using proper source branch

PR #271, opened 2025-11-21
 refactor(dashboard): remove unused default_satellite fallback

Flight-Lab:
The `default_satellite` fallback in device control variables was theoretically problematic (controls would target wrong device) but was actually **dead code** that could never execute due to system architecture.

Status icons (mic, mediaplayer, wake) are dynamically added/removed by MenuManager based on device state changes:
- Icons only appear when device entities emit state changes
- State changes require valid entity IDs from the variables
- If localStorage isn't set and variables return empty string, no entities exist to monitor
- Therefore no state changes occur, so no icons get added and the buttons never render

The fallback could never actually control a random device because the UI elements depending on these variables wouldn't exist in the first place.

## Changes
Removed `default_satellite` from variable_template
Removed `?? variables.default_satellite` from:
- `var_mic_switch`
- `var_wake_switch`
- `var_mediaplayer_device`
- `var_mediaplayer_mute`

## Result
Cleaner code that makes the localStorage requirement explicit. Behavior unchanged since fallback path was unreachable.

### DIFF
--- a/View Assist dashboard and views/dashboard/dashboard.yaml
+++ b/View Assist dashboard and views/dashboard/dashboard.yaml
@@ -1,7 +1,7 @@
 button_card_templates:
   variable_template:
     variables:
-      dashboardversion: 1.1.2
+      dashboardversion: 1.2.1
       var_assistsat_entity: |-
         [[[
           return localStorage.getItem("view_assist_sensor")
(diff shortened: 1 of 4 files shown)


dinki:
Good catch.  This is a relic from the old days for sure.

PR #272, opened 2025-11-21
refactor(dashboard): use var_assistsat_entity in device control variables

Flight-Lab:
## Current State
The device control variables (`var_mic_switch`, `var_wake_switch`, `var_mediaplayer_device`, `var_mediaplayer_mute`) all repeat the same localStorage lookup:
```javascript
var micdevice_assistbid = localStorage.getItem("view_assist_sensor");
```

This is unnecessary since `var_assistsat_entity` already provides this value.

## Changes
Refactored all device control variables to use `variables.var_assistsat_entity` instead of repeating the localStorage call. The entity ID is now retrieved once and reused everywhere.

## Benefits
- **Single source of truth**: Device identification happens in one place
- **Cleaner code**
- **Easier maintenance**
- **Same functionality**: No behavioral changes, just cleaner implementation

### DIFF
--- a/View Assist dashboard and views/dashboard/dashboard.yaml
+++ b/View Assist dashboard and views/dashboard/dashboard.yaml
@@ -8,18 +8,15 @@ button_card_templates:
         ]]]
       var_mic_switch: |-
         [[[
-          try
-          {
-            var micdevice_assistbid = localStorage.getItem("view_assist_sensor");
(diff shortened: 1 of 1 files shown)


Flight-Lab:
Waiting to mark as ready until https://github.com/dinki/View-Assist/pull/271 is merged.

PR #292, opened 2025-12-06
feat(camera): auto-switch to kitt_bar overlay on camera views

Flight-Lab:
### Description
Automatically switches to kitt_bar overlay when entering camera view and restores configured default overlay when navigating away. This is to keep the camera feed(s) visible while using Assist.

### Changes:
- camera.yaml: Added var_setup_camera_overlay variable with cleanup logic
- advancedcamera.yaml: Added var_setup_camera_overlay variable with cleanup logic
- Both views now force kitt_bar overlay on entry via _overlay_init custom field
- Overlay restoration happens automatically when var_current_view !== 'camera'

### DIFF
--- a/View Assist dashboard and views/views/camera/advancedcamera.yaml
+++ b/View Assist dashboard and views/views/camera/advancedcamera.yaml
@@ -135,6 +135,53 @@ variables:
       const timeout = variables.var_url_params.get('timeout');
       return timeout ? parseInt(timeout, 10) : 0;
     ]]]
+  var_setup_camera_overlay: |-
+    [[[
+      const cleanupOverlay = () => {
+        if (window.vaCameraOverlaySet && window.vaCameraOverlayEntity) {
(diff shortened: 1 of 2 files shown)


Flight-Lab:
No, I don't think you're overcomplicating it at all. I also realized it shouldn't change the flashing bar to the kit bar cause that is already out of the way.

As for making it optional, we absolutely could do that with an additional attribute. We could probably use an additional attribute reflecting the header config setting to fix the #294 / #297 issue as well. That begs the question, is there a limit to how many attributes we want to add, whether through the master/device configuration or through the blueprints? I've been trying to limit the amount of attributes added, but it is certainly an easy way of passing information to the front end.

Flight-Lab:
You made some good points in your last comment but I haven't gotten around to implementing that logic yet.

dinki:
Stale

PR #298, opened 2025-12-13
Added 5 extra sentences pause, resume, whats playing, skip and previous

### LINKED ISSUE #163: Add "what song is playing" to device functions
It is often desired to know what song is playing.  Add this to device functions BP so that when called it returns the artist and album if availabe on media or music player device.

Rymez2K:
Added pause, resume, what song is playing,
skip and previous functionality de and pt
translations added but these are just from
translate.
Closes #163


dinki:
Thank you for your contribution.  I have updated the `device functions` blueprint to add the "What song is playing" function.  The others appear to be part of HA core sentences now so I will not replicate those functions.

PR #300, opened 2025-12-16
feat(dashboard): Add Status Icons Support for Custom Dashboards in View Assist

Flight-Lab:
Enables users to integrate their own custom dashboards (especially sections views) into View Assist while maintaining the signature status icons display in the upper right corner, providing a consistent View Assist experience across all views.

### Changes
- **Refactored** status icons logic into `status_icons_content` shared template
- **Added** `status_icons_overlay` template for sections views
- **Updated** `body_template` to use shared template

### Usage
Add status icons to any view:
```yaml
type: custom:button-card
template:
  - status_icons_overlay
```

The status icons will appear in the upper right corner just like they do in native View Assist views.

### DIFF
--- a/View Assist dashboard and views/dashboard/dashboard.yaml
+++ b/View Assist dashboard and views/dashboard/dashboard.yaml
@@ -515,6 +515,34 @@ button_card_templates:
 
                   return buttonList;
                 }]]]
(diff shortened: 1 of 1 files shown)


dinki:
Very nice!!

PR #315, opened 2025-12-23
Updated satellite selector ind devicealerts to only show satellite sensors

hamsando:
For the satellite entity selector in the existing blue-print-devicealerts.yaml, the only filter was for devices belonging to the integration. However if there is an update to the integration, then that also shows up as a selectable satellite. For example, HA is showing me "View Assist - Get Sport Scores blueprint" as an update, and since it belongs to the View Assist integration it shows up in the list of satellites in the Blueprint. This fix adds an additional filter (domain: sensor) which is ANDed with the Integration filter to only show the true View Assist satellites.


dinki:
Thanks for catching this!

PR #321, opened 2026-01-04
Update 'Pyscript' info of 'Search Wikipedia' blueprint

DJaeger:
Clearifies and links the 'Pyscript' dependency of the 'Search Wikipedia' blueprint in docs


dinki:
Thank you for this clarification.  Very much appreciated.  Accepted.

PR #357, opened 2026-02-01
support times from google stt api

patfreeman:
Using Google's Speech to Text API the timer values are returned with hyphens in them. Examples: "5-minute" or "10-second".

Regexing out the hyphens allow the blueprint to interpret time from Google STT API, and prevents the error:
"Unable to decode time or interval information"


dinki:
Thanks for this.  Looks like this shouldn't be a problem.  Will give it a good look soon.

You may be interested in the work I've been doing with adding improvement to this BP.  Take a look at this branch for the changes.

https://github.com/dinki/View-Assist/tree/viewassist-timers/View_Assist_custom_sentences/Alarms_Reminders_Timers

PR #390, opened 2026-04-13
New Play Radio Blueprint with aliase

Raabi91:
It's the new New Play Radio Blueprint with aliases, which I've already presented on Discord. I've also updated the wiki.


dinki:
This is great!  Thanks so much for this.

PR #398, opened 2026-05-08
Update blueprint-spellaword.yaml to add option for slower spelling

pikawho1234:
Added an boolean input option to slow down the spelling of the word by changing the delimiter between the letters in the response


pikawho1234:
This is my first ever PR so let me know if I should do anything differently please.

dinki:
Thank you for the contribution!

PR #409, opened 2026-05-24
Add Italian translations for music assistant responses & little bug fix

dalailama333:
Addes Italian translation for View Assist responses.

Fix for a little bug: when a song is requested to be queued, the response was always "playing song" instead "queuing song".


dinki:
Thanks for this as well!

PR #410, opened 2026-05-24
Add Italian translations for play commands

dalailama333:
Added Italian translations for music commands for selecting song in the view assist - Play Music With Music Assistant


dinki:
Thank you for providing!

PR #425, opened 2026-06-28
Patch 9

AshaiRey:
NL Translation added


dinki:
Is this the same as #426?  If so, please close any PRs that include the same files being changed so I can review whatever is the latest?   Thanks for putting time into making VA better.

dinki:
Added view camera translation manually.  Thank you!

PR #426, opened 2026-06-28
NL Translation

AshaiRey:


dinki:
Updated NL translation for Alarms...   Thank you!

PR #428, opened 2026-06-28
Patch 2

AshaiRey:
NL Translation


dinki:
Regarding device functions, I do not know if this is an improvement or not as I do not speak the language.

PR #429, opened 2026-06-28
Patch 3

AshaiRey:
NL Translation


dinki:
Added NL to sports scores BP.  Thank you!

PR #430, opened 2026-06-28
Patch 4

AshaiRey:
NL Translation


dinki:
Added NL translation for List Management.  Thank you!

PR #431, opened 2026-06-28
Patch 5

AshaiRey:
NL Translation


dinki:
Added NL to Locate BP.  Thank you!

PR #433, opened 2026-06-28
Patch 7

AshaiRey:
NL Translation


dinki:
Added NL for Spell BP.  Thank you!

PR #434, opened 2026-06-28
Patch 8

AshaiRey:
NL Translation


dinki:
I've manually added the NL translation for Travel Times by Waze BP.  We will be replacing the Play Music BP soon so will hold off on that one.  Thanks for providing this!

PR #435, opened 2026-06-28
Patch 12

AshaiRey:
NL Translation


dinki:
Thank you for the work you have put in.  Can you tell me if all these 'Patch XX' submissions are different?  It's hard to tell without digging into each.  If they are different, please edit the title to be more descript.

AshaiRey:
Hi, yes they are all different. However, because github didn't behave well for me I posted them also as a zip on discord. Sorry I can not edit the title because I am on a long trip without a computer

dinki:
Added NL translation to conversions BP.  Thanks!

PR #436, opened 2026-07-04
Add Italian translations to core blueprints

thekoma:
Adds Italian (it) translations, reviewed by a native speaker:

- Alarms Reminders & Timers (v 1.3.2)
- Device Functions (v 1.4.7)
- Play Music with Music Assistant (v 1.2.8)
- Play Radio with Music Assistant (v 1.3.3)
- Search Wikipedia (v 1.1.5)
- Thermostat Control (v 1.1.6)
- Travel Times by Waze (v 1.0.4)
- View Camera (v 2.0.3)

Also completes the existing Italian block in What time is it (v 1.0.4): it only had `responses`, so asking for the day or date in Italian raised a template error (`translations[language]['weekdays']` missing). Added `weekdays`/`months` and switched the `date` response to the composed `{day_of_week} {day_of_month} {month} {year}` pattern (like de/pl), since `{date}` interpolates English month/day names from `strftime`.

Wiki updated accordingly (language tables + per-blueprint changelog rows). While editing those rows I also aligned a few language lists that had drifted from the blueprints (missing `pt`/`pl` in some rows; `fr` listed for Play Music which the blueprint doesn't have).

Also added Italian suggested command translations to the wiki pages that document them (Device Functions, Play Music with MA, Play Radio with MA), phrased colloquially rather than as literal translations. One caveat documented inline: the Adjust Volume action compares the `{up_down}` slot against the literal English words `up`/`down`, so translated direction words (Italian and German alike) are captured but not acted on — fixing that would need a per-language word mapping in the blueprint, which felt out of scope for a translation PR.

Note for maintainers: the `fr` and `nl` blocks in What time is it have the same missing-weekdays/months issue as Italian had — I didn't touch those since I'm not a native speaker, but happy to scaffold them if someone can review.

PR #438, opened 2026-07-04
Device Functions: split Adjust Volume into Volume Up/Down commands (fixes #437)

### LINKED ISSUE #437: Device Functions: Adjust Volume matches literal English "up"/"down", breaking translated commands
### The problem

In the Device Functions blueprint, the Adjust Volume command captures the direction word in a wildcard slot:

```yaml
adjust_volume:
  default: turn {up_down} [the] volume
```

and the action compares that slot against the literal English words (`blueprint-devicefunctions.yaml`, lines 379-384):

```yaml
- service:
    '{% if trigger.slots.up_down == "down" %} media_player.volume_down {% endif %}
    {% if trigger.slots.up_down == "up" %} media_player.volume_up {% endif %}'
```

Two consequences:

1. **Translated commands silently do nothing.** A user who configures a translated command (e.g. Italian `{up_down} il volume` → "alza il volume", or the German suggestion documented in the wiki, `Schalte [die] Lautstärke {up_down}` → "lauter") gets the direction word captured, but it never matches `"up"`/`"down"`, so the service template renders empty and no volume change happens.
2. **The response echoes the raw slot even for English users of other languages.** `volume_adjusted` does `.replace('{direction}', trigger.slots.up_down)`, so with English triggers and Italian responses you get "Il volume è stato **up**".

### Proposed fix

Use the same pattern Thermostat Control already uses for raise/lower: two separate command inputs with their own trigger ids, instead of one input with a direction slot.

- Replace the `adjust_volume` input with `volume_up_command` (default `turn up [the] volume`) and `volume_down_command` (default `turn down [the] volume`)
- Branch on trigger id: `volume_up` → `media_player.volume_up`, `volume_down` → `media_player.volume_down`
- Replace the `volume_adjusted` response with two keys, `volume_raised` / `volume_lowered`, so every language can phrase each direction naturally instead of echoing the spoken word

This makes the command fully translatable with no hidden English dependency, and fixes the response for all non-English languages at once.

**Trade-off to be aware of:** replacing the `adjust_volume` input is a breaking change for users who customized that command — they would need to re-enter their phrases in the two new inputs (existing automations keep working with the new defaults otherwise). That is why I'm opening this issue first rather than just sending the change.

### Offer

I have this implemented and ready to submit as a PR (built on top of #436), including derived `volume_raised`/`volume_lowered` strings for **all currently supported languages** (ca, de, en, es, it, nl, pl, pt, sr, sr-Latn), based on each language's existing `volume_adjusted` phrasing. The Italian ones are native-reviewed; the others are conservative derivations that would benefit from a native speaker's confirmation.

While preparing this I also noticed the `nl:` block in Device Functions currently contains German strings for `view_switch`, `view_invalid`, `volume_set`, `volume_adjusted` and `sound_restored` (looks like a copy-paste slip) — happy to flag that separately if useful.

Let me know if the approach works for you and I'll open the PR.

thekoma:
Fixes #437.

**Note: this branch is stacked on #436** (Italian translations) — merge that first and this diff reduces to a single commit touching only Device Functions.

## What changed

- Replaced the `adjust_volume` input (`turn {up_down} [the] volume`) with two inputs, `volume_up_command` / `volume_down_command`, branching on trigger id — the same pattern Thermostat Control uses for raise/lower. Translated command phrases now work with no hidden dependency on the English words `up`/`down`.
- Replaced the `volume_adjusted` response (which echoed the raw slot word, producing things like "Il volume è stato **up**") with per-direction keys `volume_raised` / `volume_lowered` in **all** supported languages, derived from each language's existing `volume_adjusted` phrasing: ca, de, en, es, it, nl, pl, pt, sr, sr-Latn. Italian is native-reviewed; the others are conservative derivations — a native speaker's once-over is welcome.
- While in there, fixed a pre-existing swap: the `de:` block was missing `view_switch`, `view_invalid`, `volume_set` and `sound_restored`, and those German strings were sitting in the `nl:` block (German users currently get template errors on those code paths). Moved them home and gave `nl:` proper Dutch equivalents (also flagged for native review).
- Wiki: split the Adjust Volume suggested-command section into Volume Up / Volume Down (en/de/it examples), changelog row, version bump to v 1.5.0.

## Breaking change

Users who customized the old `adjust_volume` command need to re-enter their phrases in the two new inputs (everyone else just picks up the new defaults). Happy to adjust the approach if you'd rather keep the old input around.

Validated: YAML parses, and all 9 non-English language blocks have exact key parity with `en:`.

### DIFF
--- a/View_Assist_custom_sentences/Device_Functions/blueprint-devicefunctions.yaml
+++ b/View_Assist_custom_sentences/Device_Functions/blueprint-devicefunctions.yaml
@@ -2,7 +2,7 @@ blueprint:
   name: View Assist - Device Functions
   description:
     Various device functions for the View Assist Satellites (View Assist
-    devicefunctions v 1.4.7)
+    devicefunctions v 1.5.0)
   domain: automation
   input:
     language:
@@ -34,10 +34,14 @@ blueprint:
       name: Set Volume Command
       description: The command to use to set volume level directly
       default: "[set | turn] [the] volume [to] {level}"
-    adjust_volume:
-      name: Adjust Volume Command
-      description: The command to use to set volume level by step
-      default: turn {up_down} [the] volume
+    volume_up_command:
+      name: Volume Up Command
+      description: The command to use to raise volume by one step
+      default: turn up [the] volume
+    volume_down_command:
+      name: Volume Down Command
+      description: The command to use to lower volume by one step
+      default: turn down [the] volume
     mute_volume:
       name: Mute Volume Command
       description: The command to use to mute audio
@@ -79,8 +83,11 @@ trigger:
     command: !input set_volume
     id: set_volume
   - trigger: conversation
-    command: !input adjust_volume
-    id: adjust_volume
+    command: !input volume_up_command
+    id: volume_up
+  - trigger: conversation
+    command: !input volume_down_command
+    id: volume_down
   - trigger: conversation
     command: !input mute_volume
     id: mute_volume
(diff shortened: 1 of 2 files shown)


msp1974:
@dinki .  This PR looks good too.  Thanks again @thekoma !

dinki:
Thanks again @thekoma !

PR #460, opened 2026-07-31
View Assist Custom Prompt Responses update

relust:
Switched the TTS call from media_player.play_media to tts.speak, since play_media had become unreliable and the media player wasn't always playing the response. Added a TTS engine selector and a voice/language field, plus a playback-aware wait before re-enabling the microphone.


dinki:
Hey @relust !  Hope you have been well.  Thanks for the update.

Issue #164, opened 2025-04-18
Add fade in/out to view transitions


dinki:
Transitions can have a more professional look if they fade out from current and in to new.  This is not difficult to code in CSS but will need a mechanism to trigger the fade out, then swap screen, then trigger the fade in.

PengBG:
> A bit off topic, but it would also be cool to "slide" into a view with touch controls (for example, sliding the main view to the right brings you to a view to control the lights in your house).

This would be almost impossible to implement for all people, but you can make your own dashboard as new dashboard in View assist and install [this](https://github.com/zanna-37/hass-swipe-navigation) on your HA. Some people would like lights, some would like climate, some would love alarm, some maybe a camera....

PR #475, opened 2026-08-29
feat: add CSS view transitions and configurable transition duration

arcsur:
### Summary
This PR implements smooth, context-aware CSS fade transitions for View Assist view changes and mode switching, addressing https://github.com/dinki/View-Assist/issues/164.

### Background & Motivation
As discussed in https://github.com/dinki/View-Assist/issues/164:
- When changing views or mode states, abrupt visual switches can be jarring.
- **Shared Backgrounds (e.g. Clock ↔ ClockAlt)**: Fading to black when switching between views that share the same satellite wallpaper (or live slideshow camera feed) creates unnecessary screen flashes. For these views, only the foreground UI components (`#container`) fade in/out while the background remains unbroken.
- **Dedicated Backgrounds (e.g. Clock ↔ Weather / Music / Info)**: When transitioning to or from views with their own custom colors or background images, the entire view card (`#card`) smoothly fades to and from black (like the Amazon Echo Show behavior) for a polished, seamless experience.

### Key Changes
- **Context-Aware CSS Transitions (`dashboard.yaml`)**:
  - Implemented dual-mode animation logic in `body_template.extra_styles`:
    - Views sharing the satellite wallpaper animate `#container` with `@keyframes vaFadeInUI` / `@keyframes vaFadeOutUI`.
    - Views transitioning to/from dedicated backgrounds animate `#card` with `@keyframes vaFadeInCard` / `@keyframes vaFadeOutCard` (fade to/from black).
  - Tracks previous view background state via `window.viewAssistPrevWasCustom` to adapt transitions bidirectionally.
  - Dynamically respects configured duration from `attributes.view_transition_time` (default `0.5s`).
  - Added default `var_custom_background: false` to `variable_template`.
  - Feature version bump `dashboardversion` to `1.6.0`.
- **View Background Declarations & Patch Version Bumps**:
  - Declared `var_custom_background: true` on views with dedicated backgrounds (`weather.yaml`, `music.yaml`, `music-alternative.yaml`, `info.yaml`, `infopic.yaml`, `alert.yaml`, `thermostat.yaml`, `sports.yaml`, `intent.yaml`, `locate.yaml`, `webpage.yaml`, `list.yaml`, `list-nocheckbox.yaml`, and community contribution `weatherdynamic.yaml`).
  - Declared `var_custom_background: false` on views sharing the satellite wallpaper (`clock.yaml`, `clockalt.yaml`, `calendar.yaml`, `alarm.yaml`, `camera.yaml`, `advancedcamera.yaml`, and community contribution `clockaltwithmovement.yaml`).
  - Patch version bumped all 21 view definitions.
- **Clock Cards Smooth In-Place Fade (`clock.yaml` & `clockalt.yaml`)**:
  - Smooth CSS `transition` for `opacity` and `color` on time and weather items between normal and night modes.
- **Documentation (`dashboard-options.md`)**:
  - Updated documentation explaining the context-aware view transition behavior between shared wallpapers and dedicated background views.

### Related Issue & Integration PR
- Resolves https://github.com/dinki/View-Assist/issues/164
- Companion Integration PR: https://github.com/dinki/view_assist_integration/pull/268

### DIFF
--- a/View Assist dashboard and views/dashboard/dashboard.yaml
+++ b/View Assist dashboard and views/dashboard/dashboard.yaml
@@ -1,7 +1,7 @@
 button_card_templates:
   variable_template:
     variables:
-      dashboardversion: 1.4.0
+      dashboardversion: 1.5.0
       var_assistsat_entity: |-
         [[[
           return localStorage.getItem("view_assist_sensor")
@@ -98,6 +98,13 @@ button_card_templates:
           try
           {
             return states[variables.var_assistsat_entity].attributes.use_24_hour_time;
+          } catch { return  false}
+        ]]]
+      var_assistsat_entity_enable_view_transitions: |
+        [[[
+          try
+          {
+            return states[variables.var_assistsat_entity].attributes.enable_view_transitions || false;
           } catch { return  false}
         ]]]
       var_title: |-
@@ -252,6 +259,42 @@ button_card_templates:
       - variable_template
       - status_icons_content
       - responsive_base
+    extra_styles: |-
+      [[[
+        try {
+          const transitionsEnabled = states[variables.var_assistsat_entity]?.attributes?.enable_view_transitions;
+          if (transitionsEnabled === true || transitionsEnabled === "true") {
+            return `
+              @keyframes vaFadeInUI {
+                0% {
+                  opacity: 0;
+                }
+                100% {
+                  opacity: 1;
+                }
+              }
+              @keyframes vaFadeOutUI {
+                0% {
+                  opacity: 1;
+                }
+                100% {
+                  opacity: 0;
+                }
+              }
+              #container {
+                animation: vaFadeInUI 0.3s ease-out forwards;
+                will-change: opacity;
+              }
+              .va-fade-out #container {
+                animation: vaFadeOutUI 0.15s ease-in forwards !important;
+              }
+            `;
+          }
+          return "";
+        } catch {
+          return "";
+        }
+      ]]]
     show_state: false
     show_icon: false
     show_name: false
--- a/View Assist dashboard and views/dashboard/dashboard.yaml
+++ b/View Assist dashboard and views/dashboard/dashboard.yaml
@@ -314,7 +314,7 @@ button_card_templates:
                   animation: vaFadeInCard ${fadeInDuration}s ease-out forwards;
                   will-change: opacity;
                 }
-                :host(.va-fade-out) #card, #card.va-fade-out, .va-fade-out #card {
+                :host(.va-fade-out) #card, #card.va-fade-out {
                   animation: vaFadeOutCard ${fadeOutDuration}s ease-in forwards !important;
                 }
               `;
@@ -340,7 +340,7 @@ button_card_templates:
                   animation: vaFadeInUI ${fadeInDuration}s ease-out forwards;
                   will-change: opacity;
                 }
-                :host(.va-fade-out) #container, #container.va-fade-out, .va-fade-out #container {
+                :host(.va-fade-out) #container, #container.va-fade-out {
                   animation: vaFadeOutUI ${fadeOutDuration}s ease-in forwards !important;
                 }
               `;
--- a/View Assist dashboard and views/dashboard/dashboard.yaml
+++ b/View Assist dashboard and views/dashboard/dashboard.yaml
@@ -1,7 +1,7 @@
 button_card_templates:
   variable_template:
     variables:
-      dashboardversion: 1.5.1
+      dashboardversion: 1.6.0
       var_assistsat_entity: |-
         [[[
           return localStorage.getItem("view_assist_sensor")
@@ -117,6 +117,7 @@ button_card_templates:
             return (t !== undefined && t !== null && !isNaN(t)) ? parseFloat(t) : 0.5;
           } catch { return  0.5}
         ]]]
+      var_custom_background: false
       var_title: |-
         [[[
           try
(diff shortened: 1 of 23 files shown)


dinki:
@arcsur Just gave this a try.  Very nice work.  I see a really smooth transition when moving from normal to night mode on the clock.

I did try switching to weather and then timeout back to clock (normal mode) and did not see the transition when going to either.  Is this not implemented yet?  I am hoping that when switching to any and all views that the effect would show.  My old Echo Show devices did this by default and it appears to be a better look than the abrupt change VA presents normally.

arcsur:
@dinki It should work.  Did you enable the setting for it?  I was a little unsure about enabling it by default for low power devices so I put a setting on it, default is OFF.

dinki:
@arcsur For testing, I modified this way:

```
      var_assistsat_entity_enable_view_transitions: |
        [[[
          try
          {
            const s = hass?.states?.[variables.var_assistsat_entity] || (typeof states !== 'undefined' ? states?.[variables.var_assistsat_entity] : null);
            return s?.attributes?.enable_view_transitions || false;
          } catch { return  true}
        ]]]
```

Setting to true should enable right?  I have not pushed to having this set in the satellite yet.

arcsur:
@dinki Yes, need to set to true as it is default false. You can also enable with a "service call"

Target Entity: sensor.web_dashboard
enable_view_transitions: true
view_transition_time: 0.5 (seconds)

dinki:
Hold tight.  I am only testing the VA portion for the dashboard and view right now.  Is this dependent on having the VA integration portion enabled to get the full feature?  I thought I was being smart in trying to skirt around that piece but maybe it is necessary?  Admittedly I haven't looked at that PR closely.

arcsur:
You don't need to update the VA integration changes if you set the attribute via the service call `enable_view_transitions: true`
The update to the integration just allows turning on the CSS integration, or setting the transition time (default is 0.5 seconds which is quite quick, increase to 1 second f you really want to see it. `view_transition_time: 1`

arcsur:
Alternatively, can install the integration PR, that adds the settings: https://github.com/dinki/view_assist_integration/pull/268

dinki:
Yeah I thought that's what my dashboard change was doing but I maybe not.  I'll give it another look.

dinki:
Ahhh okay.  Thanks for the hint about the time.  I am able to see what is happening now.  So this is working for the elements fading in and out but not the background.   My original idea was the mimic what the Amazon device does in that it fades to black and then fades out from black to the content.  Seems you are excluding the background.

Any thoughts on modifying it?  This may be somewhat difficult to get the timing correct as the view is changing and displaying the new background at the same time the fading is happening if that makes sense.  A delay may be necessary in the view switch so that it only switches when the screen is all the way to black.

dinki:
See this video for an example.  It shows within the first ten seconds of the video:

https://www.youtube.com/watch?v=EbgzhH1yRyw

arcsur:
So I specifically, did not want to change the background.  This is due to the fact that you can have custom backgrounds set and these should not IMO fade in and out.  In my particular case, I have an automation that takes an album slideshow camera and set the background image when it changes.  I don't want to fade the same images in and out, only the content on top.  For the default weather, it also does not matter as it renders the widget fading in/out on top of the background.  I am not so sure about advanced weather though.

All in all, I think fading the content in/out and leaving the background alone is the better approach.  It keeps a separation between the content and the background and avoids having to cross fade or fade in/out of black.

arcsur:
> See this video for an example. It shows within the first ten seconds of the video:
> 
> https://www.youtube.com/watch?v=EbgzhH1yRyw

This video demonstrates the entire screen fading to black and then fading from black to the new screen within about 0.5s.  This works for screens with differing backgrounds but if you have a fairly consistent background then not so much.

arcsur:
I have taken another look at this on my screens,  so, it really depends on whether we are using a view with it's own background vs a view that uses the shared background.

I think it is possible to solve for this, and cater for both, but it is likely going to need a more significant change.  I need to give it some thought.

arcsur:
@dinki I think I managed to get the best of both worlds.

1. If fading between 2 views that use the shared common background image, only fade in/out the components on top and do not fade the shared background (same as current)
2. If fading to/from a view which has it's own background, use a fade to/from black to make the fade smoother.

This works but I needed to add a variable to all the views to identify whether or not they have a custom background or use the shared background.  This defaults to false (shared background).  Now there is alsom fallback mechanism that tries to detect a custom background in use but most views don't defice variables.background (except info/infopic/list/list-checkbox).  And some like the weatherdynamic one computes it's own backgrounds.  Therefore, I think adding the variable is the best approach.

arcsur:
@dinki I also added fade outs.  This is now identical to the way Amazon show does it.  However, it only fades out when navigating using views_assist actions or browser_navigate().  It won't work if you directly load in lovelace for testing.

Fade in works regardless and is purely CSS based.  Fade out required some minor changes to view_assist.js in the integration.  See the related PR https://github.com/dinki/view_assist_integration/pull/268

arcsur:
Couple more minor bug fixes were added.  This change is unfortunately a lot more complex to review than prior. However, it supports your use case and mine so it may be worth it.

dinki:
I haven't forgotten about this but have not had a chance to test yet.

dinki:
hey @arcsur finally getting to check this.  I pinged you on discord but I THINK I'm testing this correctly.  Checked out the ePR for the integration and modified my dashboard, clock and weather view, enabled transitions in VA config.  I will record a video to show you what I am seeing but to describe it it's like the clock starts to fade to black, clock shows again, black again then weather shows up.  The weather goes black, then visible, then black again.  It's like I'm seeing the black screen twice for each transition.  This is on an Echo Show 5 so maybe it has something to do with timing?

Will capture video for you when I get back to my office

arcsur:
Make sure you have updated the javascript from the integration PR for fade outs to work correctly.  The javascript is cached so you need to version bump it in const.py for cache busting.

https://github.com/dinki/view_assist_integration/pull/268

PR #481, opened 2026-09-10
Fix ROM link for Echo Show 5 Gen 2 (cronos) on supported devices wiki page

iamjoshk:
Updated the ROM installation link for Echo Show 5 Gen 2 (cronos). 
Fixing the link from checkers (Echo Show 5 gen 1) to cronos for the Echo Show 5 Gen 2.


dinki:
Thank you for this update!

PR #482, opened 2026-09-11
Add configurable background color to alert view via var_background

opsec-infosec:
# Alerts Page Background Color

## Summary
I wanted a way to change the alerts background color for specific automations, such as an alarm system tripping, so I can display a red background for example, or maybe yellow if it arming.  I though it could be useful for others

## Modification

Adds alert_data['background'] as an optional attribute for the alert
view. Falls back to the existing default (#059bf1) when unset, so
existing automations/blueprints that don't set this key are unaffected.
The background attribute expects valid CSS color strings and has no validation.
Also cleaned up some white spaces, sorry my VS Code automatically does this to keep code clean


dinki:
Any chance you can provide a visual (images or video) of this in action?

